PR #1, opened 2015-06-02
Move course listing features to rspec

derekprior:
I'll be adding authorization functionality here and wanted these to be
in RSpec as I did it.


mrgilman:
:+1:

PR #4, opened 2015-06-03
Add authorization with Pundit

derekprior:
Previously, only the home page had any sort of authorization associated to it. This change introduces Pundit so we can start authorizing all actions based on the roles the user has the access required for the action. I have not added unit tests for the policies as of yet, as the sole policy is sufficiently tested via feature and controller specs.

This also introduces a `PermissionSet` object which allows us to ask questions of the entire set of permissions in an easy way.


app/models/permission_set.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    new(permissions)
+  end
+
+  def initialize(permissions)
+    @permissions = Array(permissions)
+  end
+
+  def department_slugs
+    inject([]) do |slugs, permission|
```
mrgilman: I *think* that you could use `each_with_object` instead of `inject` and you wouldn't need to return `slugs`.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,59 @@
...
+    @user = user
+    @record = record
+  end
+
+  def index?
+    false
+  end
+
+  def show?
+    scope.where(:id => record.id).exists?
```
mrgilman: Will this return a `401` if a record doesn't exist?

mrgilman:
LGTM

app/models/permission_set.rb
derekprior: good call. Updated.

PR #5, opened 2015-06-03
Rename department.role_department to slug

derekprior:
This field maps to the department name from the roles database. It is a
unique, string representation of the department. We typically refer to
these as slugs. `slug` is less redundant, shorter, and I believe less
confusing than `role_department`.

I have also added a unique index to ensure that we don't duplicate these
values and make them slightly quicker to search for.


mrgilman:
:ship: :it:

PR #6, opened 2015-06-03
Change course schema

derekprior:
* Rename `course_name` to `name` to avoid the common
  `course.course_name` redundancy
* Rename `course_number` to `number` for similar reasons.
* Add `null: false` constraints to appropriate fields
* Make `course.number` unique
* Add a foreign key from courses to departments. The foreign key is set
  to prevent deleting departments that have courses. We can revisit this
  if need be.


mrgilman:
:+1:

PR #7, opened 2015-06-03
Add timestamps to previously existing tables

derekprior:
It's a general best practice to have timestamps be present on all
tables. This tends to be helpful somewhere along the way in auditing or
debugging.


mrgilman:
:tada:

PR #8, opened 2015-06-03
Display alignments with standard outcomes on outcomes#show

mrgilman:
* Replace show_assessments.feature with RSpec feature specs
* Add link to create new custom outcomes from courses#show


app/views/outcomes/show.html.erb
```diff
@@ -1,4 +1,21 @@
-<%=@outcome.name%>:&nbsp;<%=@outcome.description %>
+<h3><%=@outcome.name%>:&nbsp;<%=@outcome.description %></h3>
```
derekprior: do you need the `&nbsp;` here?

derekprior:
One tiny comment. :+1:

app/views/outcomes/show.html.erb
mrgilman: Unlikely. There are a few of these around so I think I'll submit a PR after this one to clean them all up.

PR #9, opened 2015-06-03
Extract RolesDB to external repo

derekprior:
This previously existed vendored in the app, which can lead to version
safety issues and also makes it less discoverable. I believe the
original code was written by Cory Foy, though I'm not certain and I
can't track down any original repositories for it. The commit history
for the vendored gem yielded no good details on this.


mrgilman:
:+1:

PR #10, opened 2015-06-03
Remove dedicated interstitial action

derekprior:
This can more RESTfully be achieved with a conditional in the view rather
than a conditional redirect in the controller. This will make
permissioning for this action simpler as well.

This also moves the "adopt standard outcomes" tests to rspec.


mrgilman:
:clap:

PR #12, opened 2015-06-03
Use STI for indirect assessment types

mrgilman:
* Create separate models, controllers, and forms for each indirect assessment type
* Minimal change to schema -- only requires addition of `type` column to indirect assessments
* Deleted the last of the cucumber specs! :eggplant:


app/models/outcome.rb
```diff
@@ -2,6 +2,9 @@ class Outcome < ActiveRecord::Base
   belongs_to :course
   has_many :direct_assessments
   has_many :indirect_assessments
```
derekprior: Can this go away now?

derekprior: Oh wait, nevermind...

config/routes.rb
```diff
@@ -15,6 +15,9 @@
     resources :assessments
     resources :direct_assessments
     resources :indirect_assessments
```
derekprior: Can this go away?

derekprior:
Cool. Quick win via STI - this is a a nice improvement.

config/routes.rb
mrgilman: It's in use on the outcomes#show page to display the table of indirect assessments.

derekprior: :+1:

PR #13, opened 2015-06-03
Add authorization on course show

derekprior:
Authorization is driven by the department level access. If you have read
only access to the department for the course, then you can view the
course.

I had to update the back door code for tests so that the supplied user
carries over to additional requests after the initial visit call. This
is done by storing the user id in the session.


mrgilman:
:star2:

PR #14, opened 2015-06-04
Make adopting custom outcomes a RESTful action

derekprior:
We can think of default outcomes as a singular resource that can be
created for a given course.

I removed the existing controller specs becuse they were adequately
covered by feature specs already.


mrgilman:
:boom:

PR #15, opened 2015-06-04
Display indirect assessment type on outcomes#show

mrgilman:
Also fix incorrectly named spec files

### DIFF
--- a/app/views/outcomes/show.html.erb
+++ b/app/views/outcomes/show.html.erb
@@ -58,6 +58,7 @@
       <th>Minimum Category</th>
       <th>Target Percentage</th>
       <th>Actual Percentage</th>
+      <th>Type</th>
     </tr>
     <% @indirect_assessments.each do |assessment| %>
       <tr id="indirect_assessment-<%=assessment.id %>">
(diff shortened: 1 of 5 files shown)


app/views/outcomes/show.html.erb
```diff
@@ -67,6 +68,7 @@
         <td><%= assessment.minimum_category%></td>
         <td><%= assessment.target_percentage%></td>
         <td><%= assessment.actual_percentage%></td>
+        <td><%= assessment_type(assessment) %></td>
```
derekprior: This would be a bit more round-about, but what do think of `assessment.class.model_name.human`. Then you can control the string with i18n:

```yml
en:
  activerecord:
    models:
      other_assessment: "Other"
      participation: "Program Participation"
```

Definitely more indirection, but that would provide consistent naming in any other place (like error message). The helper is more direct though. Your call.

derekprior:
:+1:

app/views/outcomes/show.html.erb
mrgilman: I didn't know you could do this! Though I think I might leave the helper for now and revisit if I end up needing this elsewhere.

PR #16, opened 2015-06-04
Add authorization for outcome actions

derekprior:
Also added a new helper for use in tests to ease creating users with
proper access.


mrgilman:
:ok_woman:

PR #17, opened 2015-06-04
Enable editing of indirect assessments

mrgilman:
https://trello.com/c/RACWKdgd

### DIFF
--- a/app/controllers/participations_controller.rb
+++ b/app/controllers/participations_controller.rb
@@ -16,6 +16,24 @@ class ParticipationsController < ApplicationController
     end
   end
 
+  def edit
+    @outcome = Outcome.find(params[:outcome_id])
+    @participation = Participation.find(params[:id])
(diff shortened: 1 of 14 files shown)


app/controllers/participations_controller.rb
```diff
@@ -16,6 +16,24 @@ def create
     end
   end
 
+  def edit
+    @outcome = Outcome.find(params[:outcome_id])
+    @participation = Participation.find(params[:id])
+    @available_years = ['2014', '2015', '2016', '2017', '2018', '2019']
```
derekprior: Can you think of a good way to eliminate this duplication? Could it be a view helper instead of an ivar? Maybe even calculate it in the helper so it doesn't have to be continually updated?

derekprior:
There's a lot of controller duplication here, but that's okay, I think. It's boilerplate-like duplication. Just wondering if we can clean up the years bit since that feels less like boilerplate.

derekprior:
:camel:

PR #18, opened 2015-06-04
Clean up tests

derekprior:
* Use user creation helper everywhere
* Randomize test order
* Miscellaneous changes to spec and rails helpers from suspenders


mrgilman:
:dromedary_camel:

PR #19, opened 2015-06-04
Add authorization to assessments

derekprior:
I used a concern here to map all assessment types to the same assessment
policy. Having completed all of the policies now, I'm concerned that
there's a lot of demeter violations to get back to the department which
is what we are always authorizing on. I may investigate ways to ease
these demeter violations, but I think it would require more
associations. Not sure that's a good trade off.


mrgilman:
:turtle:

PR #20, opened 2015-06-04
Centralize authorization verification

derekprior:
Now that all of the controllers have appropriate authorization calls, we
can move the filters into the application controller. This will ensure
that any new controllers require authorization to be set up properly or
intentionally skipped with `skip_after_filter`.


mrgilman:
:pig_nose:

PR #21, opened 2015-06-04
Add ability to clone assessments

mrgilman:
When a user clicks "Clone" next to a direct assessment, they are brought to a form that is prefilled with the attributes of that assessment (except for the actual percentage) so that they can easily create a new assessment that defaults to the attributes of the cloned one.

### DIFF
--- a/app/controllers/direct_assessments_controller.rb
+++ b/app/controllers/direct_assessments_controller.rb
@@ -3,7 +3,7 @@ class DirectAssessmentsController < ApplicationController
 
   def new
     @outcome = Outcome.find(params[:outcome_id])
-    @assessment = @outcome.direct_assessments.build
+    @assessment = @outcome.direct_assessments.build(new_assessment_attributes)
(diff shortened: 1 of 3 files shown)


app/controllers/direct_assessments_controller.rb
```diff
@@ -2,6 +2,9 @@ class DirectAssessmentsController < ApplicationController
   def new
     @outcome = Outcome.find(params[:outcome_id])
     @assessment = DirectAssessment.new
```
derekprior: how about:

```
DirectAssessment.new(new_assessment_attributes)
```

Then the conditional on the next line can move into the attributes method and either return an empty hash or the attributes you want to clone. There's something about a conditional that makes me want to have it be isolated to an entire method rather than inside a method. Don't feel super strongly though


The difference is pretty minor, but I think I'd prefer the

derekprior:
:+1:

PR #22, opened 2015-06-04
Allow cloning of indirect assessments

mrgilman:
Use a concern to avoid duplication of cloning logic in controllers.

I's like to be able to use the concern for direct assessments as well, but I couldn't think of a clean way of removing the model from the concern.

It would also be nice if there were a way to reduce duplication in the feature specs, like shared examples but for features.


derekprior:
:dragon_face:

PR #23, opened 2015-06-05
Render a 401 error on unauthorized requests

derekprior:
Previously, unauthorized requests resulted in a 500 error. This change
turns those exceptions into 401 errors.


mrgilman:
:+1:

PR #24, opened 2015-06-05
Don't show links to unauthorized actions

derekprior:
We shouldn't' show the user links to actions we know they are not
authorized to perform.

In the process of implementing the view specs for this, I came across a
tricky issue with regards to stubbing the `policy` method in the view
(which is provided by pundit via a controller helper method). The
`verify_partial_doubles` setting in rspec was raising an error because
the view itself does not actually implement `policy`. Rails magic!

I thought I could trick it by passing a lambda named `policy` to the
view as a local variable but that did not work. Instead, I followed some
workarounds found here: https://github.com/rspec/rspec-rails/issues/1076

Essentially in each view spec, we extend the view object to contain the
controller helper methods that we encounter in our views. Then, since
they are now technically present on the view object, we can stub them.
I'm not entirely comfortable with that solution since we're actually
changing the object under test, but I thought it was the best option.
The other things I considered were disabling partial double verification
in view specs with an around filter or testing these with feature specs.

### DIFF
--- a/app/views/courses/_interstitial.html.erb
+++ b/app/views/courses/_interstitial.html.erb
@@ -1,11 +1,18 @@
-<div class="row">
-  <h3>To set up outcomes</h3>
-</div>
-<div class="row">
-  <%=link_to "Adopt Default Outcomes",
-    course_default_outcomes_path(course),
-    method: :post %>
-</div>
-<div class="row">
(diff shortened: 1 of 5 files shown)


mrgilman:
That is wild. This looks like a good solution. :sheep:

PR #25, opened 2015-06-08
Clean up Demeter violations with delegation

derekprior:
By using delegation, we can avoid jumping through method chains, which
can be brittle. Delegation means that if the structure changes we can
update it in one place rather than spread out through the application.

### DIFF
--- a/app/models/direct_assessment.rb
+++ b/app/models/direct_assessment.rb
@@ -1,3 +1,5 @@
 class DirectAssessment < ActiveRecord::Base
   belongs_to :outcome
+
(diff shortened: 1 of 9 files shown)


mrgilman:
This looks a lot better. Thoughts on adding shoulda-matchers to test delegation directly? Or do you think they're sufficiently covered by feature specs?

mrgilman:
:panda_face:

derekprior:
Enough stuff breaks when I remove the delegation, so I'm okay with it.

mrgilman:
:rainbow:

PR #27, opened 2015-06-08
Update design to use bourbon

derekprior:
By using bourbon, neat, bitters, and refills, we can get a reasonable
looking design without having to add a number of non-semantic classes.

This change loses the breadcrumb functionality of the previous layout,
but this can be added back in a more robust manner if desired. User
testing seemed to to show people weren't finding the breadcrumbs very
useful.

I also added headings to some pages and extracted translations. I have
focused on the home page, department page, and the listing of course
objectives thus far. We can move the designs over in this direction

We will need to update the design of other parts of the application as
we touch them for the task-based workflow changes we're making.


derekprior:
outcomes-bourbon

app/assets/javascripts/navigation.js
```diff
@@ -0,0 +1,13 @@
+$(document).ready(function() {
```
derekprior: Oh, this snippet probably needs to change because turbolinks

app/views/outcomes/_outcome.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= content_tag_for(:tr, outcome) do %>
```
mrgilman: OMG. I didn't know this existed.

mrgilman:
:rabbit2:

PR #28, opened 2015-06-08
Separate out Result model from assessments

mrgilman:
https://trello.com/c/g8n3oU00

* Add separate model and form for tracking assessment results
* Create polymorphic relationship between results and all assessment
  types
* Add direct_assessments#show page which displays all associated results

### DIFF
--- a/app/policies/result_policy.rb
+++ b/app/policies/result_policy.rb
@@ -1,0 +1,5 @@
+class ResultPolicy < ApplicationPolicy
+  def create?
+    user.admin?(record.department)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,11 +10,15 @@ Rails.application.routes.draw do
 
   resources :outcomes do
     resources :assessments
-    resources :direct_assessments
+    resources :direct_assessments, only: [:new, :create]
     resources :indirect_assessments
     resources :surveys
     resources :participations
     resources :other_assessments
     resources :outcome_alignments
+  end
+
+  resources :direct_assessments, only: [:show, :edit, :update] do
+    resources :results, only: [:new, :create]
   end
 end
--- a/db/migrate/20150608190453_create_results.rb
+++ b/db/migrate/20150608190453_create_results.rb
@@ -1,0 +1,20 @@
+class CreateResults < ActiveRecord::Migration
+  def change
+    create_table :results do |t|
+      t.integer :assessment_id, null: false
+      t.string :assessment_type, null: false
+      t.string :assessment_name, null: false
+      t.string :assessment_description, null: false
+      t.string :problem_description
+      t.integer :percentage, null: false
+      t.integer :year, null: false
+      t.string :semester
+      t.timestamps null: false
+    end
+
+    add_index :results,
+      [:assessment_id, :assessment_type, :year, :semester],
+      unique: true,
+      name: "index_results_on_assessment_and_year_and_semester"
+  end
+end
--- a/app/controllers/direct_assessments_controller.rb
+++ b/app/controllers/direct_assessments_controller.rb
@@ -21,23 +21,26 @@ class DirectAssessmentsController < ApplicationController
   end
 
   def edit
-    @outcome = Outcome.find(params[:outcome_id])
     @assessment = DirectAssessment.find(params[:id])
     @available_semesters = ['2015FA', '2015JA', '2015SP']
     authorize(@assessment)
   end
 
   def update
-    @outcome = Outcome.find(params[:outcome_id])
     @assessment = DirectAssessment.find(params[:id])
     authorize(@assessment)
 
     @assessment.assign_attributes(direct_assessment_params)
     if @assessment.save
-      redirect_to outcome_path(@outcome)
+      redirect_to outcome_path(@assessment.outcome)
     else
       render :edit
     end
+  end
+
+  def show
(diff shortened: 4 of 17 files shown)


app/policies/result_policy.rb
```diff
@@ -0,0 +1,9 @@
+class ResultPolicy < ApplicationPolicy
+  def new?
```
derekprior: If you take a look at application policy, you'll see that new is defined as `create?`. If you define `create?` you will get `new?` for free.

mrgilman: Oh, neat. I guess I missed that because I needed to define `new?` first.

config/routes.rb
```diff
@@ -10,11 +10,15 @@
...
-    resources :direct_assessments
+    resources :direct_assessments, shallow: true
     resources :indirect_assessments
     resources :surveys
     resources :participations
     resources :other_assessments
     resources :outcome_alignments
   end
+
+  resources :direct_assessments, only: [] do
```
derekprior: Is this empty array here kind of a lie? Does the `shallow: true` above mean we actually have `:show`, `:edit`, `:update`, and `:destroy` here? Or does this empty array override those routes that `shallow` is providing?

Since we have to add the `direct_assessments` resource anyway (so we can nest results), maybe it'd be less confusing to be explicit about the actions in each case and ditch the `shallow`?

db/migrate/20150608190453_create_results.rb
```diff
@@ -0,0 +1,17 @@
+class CreateResults < ActiveRecord::Migration
+  def change
+    create_table :results do |t|
```
derekprior: Should some of these be null false?

config/routes.rb
mrgilman: You're right, this probably isn't a good use case for shallow routes. Going to ditch the `shallow: true` but keep the idea.

db/migrate/20150608190453_create_results.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string :name
+      t.string :description
+      t.string :problem_description
+      t.integer :percentage
+      t.integer :year
+      t.string :semester
+      t.timestamps null: false
+    end
+
+    add_index :results, [:assessment_id, :assessment_type]
```
derekprior: What do you think about adding:

```ruby
add_index :results, [:assessment_id, :assessment_type, :year, :semester], unique: true
```

The intention being to guard against duplicate results for the same period.

db/migrate/20150608190453_create_results.rb
mrgilman: Probably. Is it time to add model validations and simple_form?

db/migrate/20150608190453_create_results.rb
```diff
@@ -0,0 +1,17 @@
+class CreateResults < ActiveRecord::Migration
+  def change
+    create_table :results do |t|
+      t.integer :assessment_id
+      t.string :assessment_type
+      t.string :name
+      t.string :description
+      t.string :problem_description
```
derekprior: Should all three of these fields (name, description, problem description) editable in the result? What are some actual examples of what these values would be? Any chance we can find better names than `description` and `problem_description`? I'm just not sure how they are supposed to differ.

db/migrate/20150608190453_create_results.rb
derekprior: Also, anything we should remove from `assessments`?

db/migrate/20150608190453_create_results.rb
derekprior: I think we can focus on the data model validations only for now. And circle back around to the user facing stuff on that other card we already have.

derekprior: by "data model validations" I mean "database constraints."

db/migrate/20150608190453_create_results.rb
mrgilman: Yeah, I struggled with this a bit myself. These are the fields currently in the `direct_assessments` table that need to be editable by TA if, for example, the problem number for the assessment changes. `name` and `description` have different prefixes in the `direct_assessments` and `indirect_assessments` models (`assignment_` and `assessment_`, respectively). Would it be better to add the `assessment` prefix to those fields?

`problem_description` is only in the `direct_assessments` table, and I think it just adds more detail about the assessment.

db/migrate/20150608190453_create_results.rb
mrgilman: Yeah, I was going to do that after I wired up the new result stuff with the indirect assessments. I'll add a Trello card for it.

db/migrate/20150608190453_create_results.rb
mrgilman: :+1:

db/migrate/20150608190453_create_results.rb
jdaries: If it helps at all, :survey_question and :problem_description are analogous in the two models

derekprior:
looks good. Should we remove actual_percentage from assessment?

jdaries:
That makes sense since it will be part of Results

mrgilman:
Yeah, planning to do some cleanup and remove year/semester too once I get results working with indirect assessments, I have a trello card: https://trello.com/c/jD98Ei6p

derekprior:
:+1:

PR #29, opened 2015-06-09
Introduce Subject model

derekprior:
Direct Assessments are assigned to a subject, which allows TAs or other
people tasked with entering data from a course to quickly find the
assessments they should complete without having to know too much about
the entire system.

As part of this change, cloning assessments was removed because that
functionality is now replaced by the combination of associating a
subject to an assessment and storing assessment results (actual
percentages) in a seperate model (completed in an earlier change).

This includes a subjects importer. Jon Dairies provided the subject CSV
that this is based off of. The importer loads all subjects from the CSV
that map to departments that we are tracking in the ABET system. The
importer is run on `rake db:seed` and can also be run manually with
`rake subjects:import`. It is idempotent, so it is safe to run multiple
times and may be when new departments are added.

Currently, the assessment form has an unwieldy select box of all
courses. This will likely be updated at some point to allow for some
sort of autocomplete behavior.


mrgilman:
:koala:

PR #30, opened 2015-06-09
Consolidate controllers for indirect assessments

mrgilman:
* Have one controller for all indirect assessment types, maintaining a
  separate form partial for each type.

* Override `model_name` in STI models in order to generate urls with
  `indirect_assessment` instead of the STI model name.


derekprior:
:+1:

PR #31, opened 2015-06-09
Enable creation of results for indirect assessments

mrgilman:
* Remove prefixes on name and description fields in the
  direct_assessments and indirect_assessments tables

* Remove columns from direct and indirect assessments that are now in
* the results table

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -1,17 +1,17 @@
 class ResultsController < ApplicationController
   def new
-    @assessment = DirectAssessment.find(params[:direct_assessment_id])
+    @assessment = find_assessment
     @result = @assessment.results.build
     authorize(@result)
   end
 
   def create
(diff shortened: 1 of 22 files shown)


app/controllers/results_controller.rb
```diff
@@ -1,17 +1,17 @@
 class ResultsController < ApplicationController
+  before_action :find_assessment
```
derekprior: What do you think about calling it from the body of each action instead? It's less dry, but I tend to like seeing the instance variables that I intend to use in the view defined entirely in the controller actions.

so it'd be:

```ruby
def new
  @assessment = find_assessment
  #...
end

def create
  @assessment = find_assessment
  #...
end
```

I can go either way on this, particularly in a small controller like this.

mrgilman: Sure, I think I like this better too.

derekprior:
:+1:

PR #32, opened 2015-06-09
Add tab navigation

derekprior:
This gives us the UI entry points for the three main workflows we
discussed. However, the navigation and active tracking is not yet
functional. We still need to decide how to track active tab and provide
entry points for those actions.


derekprior:
abet_outcomes_assessment

mrgilman:
:palm_tree:

PR #33, opened 2015-06-09
Pre-fill result form with assessment attributes

mrgilman:
Name, description, and problem description should inherit from the
parents and also be editable by the TA.


derekprior:
:+1:

derekprior:
One small thing - yoou could assert the defaults (in the form values) without submitting the form in your tests. Your call.

derekprior:
:+1:

PR #35, opened 2015-06-10
Create about page

derekprior:
* Override default high voltage controller to allow for skipping pundit
  authorization on static page actions.
* Opt static pages out of the tab bar as they don't fit into this
  workflow.


derekprior:
@jdaries once merged, you can edit the `app/views/pages/about.html.erb` page to setup the content however you would like.

mrgilman:
:ok_woman:

PR #36, opened 2015-06-10
Raise on unpermitted params in dev and test

derekprior:
This is a typical part of our setup from [suspenders] and helps us catch
missing permitted parameters early. Adding this configuration showed us
a couple of places where we were not permitting the right attributes.

### DIFF
--- a/app/controllers/indirect_assessments_controller.rb
+++ b/app/controllers/indirect_assessments_controller.rb
@@ -47,7 +47,9 @@ class IndirectAssessmentsController < ApplicationController
     params.require(:indirect_assessment).permit(
       :actual_percentage,
       :description,
+      :minimum_category,
       :name,
+      :survey_question,
       :target_percentage,
       :type,
       :year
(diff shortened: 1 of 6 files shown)


app/controllers/indirect_assessments_controller.rb
```diff
@@ -47,7 +47,9 @@ def assessment_params
     params.require(:indirect_assessment).permit(
       :actual_percentage,
       :description,
+      :minimum_category,
       :name,
+      :survey_question ,
```
mrgilman: extra space

config/environments/development.rb
```diff
@@ -35,6 +35,8 @@
   # Raises error for missing translations
   config.action_view.raise_on_missing_translations = true
 
+  # Raises on unpermitted params
+  config.action_controller.action_on_unpermitted_parameters = :raise
```
mrgilman: I have been missing this. Thanks! :ok_hand:

mrgilman:
:tada:

PR #38, opened 2015-06-10
Centralize RolesDB config

derekprior:
Roles DB configuration was still present in the application
configuration when most of it had been moved to an initializer. I moved
it all to the initializer.

### DIFF
--- a/config/initializers/roles_db.rb
+++ b/config/initializers/roles_db.rb
@@ -13,3 +13,12 @@ if Rails.env.test?
     config.strategy_class = "RolesDb::LocalRoles"
   end
 end
+
+RolesDb.configure do |config|
+  config.ssl_cert_key_file = Rails.root.join("config", "certs", "outcomes-key.pem")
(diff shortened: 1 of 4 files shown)


config/initializers/roles_db.rb
```diff
@@ -13,3 +13,12 @@
     config.strategy_class = "RolesDb::LocalRoles"
   end
 end
+
+RolesDb.configure do |config|
+  config.ssl_cert_key_file = Rails.root.join("config", "certs", "outcomes-key.pem")
+  config.ssl_cert_file = Rails.root.join("config", "certs", "outcomes.cer")
+  config.ssl_ca_cert_file = Rails.root.join("config", "certs", "mit.pem")
+  config.roles_wsdl_file = Rails.root.join("config", "certs", "roles.wsdl")
+  config.ssl_cert_key_password = "abetiscool"
```
mrgilman: Is this something that we shouldn't have on public GitHub?

derekprior: I don't think so. I was going to follow up with Jon about scrubbing it. But for now, it's just moved.

PR #39, opened 2015-06-10
Remove reference to modernizr

derekprior:
This file doesn't actually exist on the file system and is not needed
any longer anyway. No need to try to precompile it.


mrgilman:
:ghost:

PR #41, opened 2015-06-10
Clean up routes

derekprior:
* Restrict resourceful actions to the routes we are using
* Make course new/create depend on parent course resource
* Organize the routes file a bit (alphabetized, wrap long lines, concern
  defined at the top)


mrgilman:
:fireworks:

PR #44, opened 2015-06-10
Add department number to departments and subjects

mrgilman:
Update seed file and subject importer to populate new columns.

### DIFF
--- a/lib/subject_importer.rb
+++ b/lib/subject_importer.rb
@@ -7,9 +7,10 @@ class SubjectImporter
 
   def run
     CSV.foreach(path, headers: true).each do |row|
-      if targeted_department_numbers.include?(row["Course Number"].to_s)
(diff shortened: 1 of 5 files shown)


lib/subject_importer.rb
```diff
@@ -20,8 +21,6 @@ def run
   attr_reader :path
 
   def targeted_department_numbers
-    @targeted_department_numbers ||= Course.pluck(:number).map do |number|
-      number.split("-").first
-    end.uniq
+    @targeted_department_numbers ||= Department.pluck(:number).map(&:to_s)
```
derekprior: Do we need both the to_s here and on line 10 or can we remove both?

mrgilman: Removed `to_s` on department number and added `to_i` on line 10 -- this will catch any courses with a letter duffix.

PR #45, opened 2015-06-10
Add outcome management dashboard

derekprior:
This is the entry point for the outcome management (admin) workflow.
It's focused on letting the user know what they might have to do next
and a little bit of the why (content to be edited by Jon later).

To facilitate listing courses that do not have outcomes, I added a
counter cache to courses so they always know how many outcomes are
associated.

Determining courses that have custom outcomes that are not fully aligned
is handled via N+1 query at the moment. This can be addressed in the
future, but most people will have access only to a few courses so it's
not the biggest issue presently.


app/models/outcomes_dashboard.rb
```diff
@@ -0,0 +1,19 @@
...
+    @courses = courses
+  end
+
+  def courses_without_outcomes
+    courses.where(outcomes_count: 0)
+  end
+
+  def unaligned_courses
+    @unaligned_courses ||= courses.with_custom_outcomes.select do |course|
+      StandardOutcome.unaligned_with(course).present?
```
mrgilman: If you use `any?` instead of `present?`, ActiveRecord magic will turn this into a `count` instead of `select *`.

spec/models/outcomes_dashboard_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      create(:outcome)
+
+      dashboard = OutcomesDashboard.new(Course)
+
+      expect(dashboard.courses_without_outcomes).to eq [course]
+    end
+  end
+
+  describe "#unaligned_courses" do
+    it "returns courses that are not full aligned" do
```
mrgilman: fully

mrgilman:
:musical_keyboard:

PR #46, opened 2015-06-10
Add current tab tracking

derekprior:
* Each top level view that wants to be displayed under tabs needs to
  declare which tab it is under by using the `tab` helper.
* The tabs are no longer displayed if there is no current tab.


mrgilman:
:dolphin:

PR #48, opened 2015-06-11
Add TA workflow for adding results

mrgilman:
* Add subjects#index page listing all subjects in permitted departments
* Add subjects#show page listing all direct assessments for a given subject,
  connecting to existing workflow for adding results


derekprior:
looks good!

PR #49, opened 2015-06-11
Refactor outcome creation

derekprior:
* Use accepts_nested_attributes to handle the alignment creation rather
  than handling it manually.
* Perform the multiple creation steps in a transaction to ensure
  integrity if one step fails.
* Use Simple Form to build the forms with less markup


app/controllers/outcomes_controller.rb
```diff
@@ -7,37 +7,48 @@ def show
...
+  def course
+    @course ||= Course.find(params[:course_id])
+  end
+
+  def create_outcome(outcome)
+    ActiveRecord::Base.transaction do
+      authorize(outcome)
+      outcome.course.adopt_custom_outcomes!
+      outcome.save!
+      true
```
mrgilman: Do you need this line if `outcome.save!` returns `true`?

mrgilman:
:volcano:

mrgilman:
(That's a good thing)

app/controllers/outcomes_controller.rb
derekprior: Good call.

PR #50, opened 2015-06-11
Add flash messages on success

derekprior:
In testing, users reported not being sure that something had completed
successfully. Flash messages are a good way to communicate this
information. We already had the styling, we just had to populate and
render them.


mrgilman:
:poodle:

PR #51, opened 2015-06-11
Move outcome management pages to outcomes tab

derekprior:
These are all pages you would be likely to view in the context of
managing outcomes.


mrgilman:
Neato. :whale:

PR #53, opened 2015-06-11
Fix intermittently failing specs

mrgilman:
When a label is generated more than 26 times (the number of letters in
the range provided to the label sequence) it starts creating outcomes
without names, which causes two specs to fail when more than 26 outcomes
have been created before the specs run. This change extends the size of
the range to 18278 so that should not happen anymore.


derekprior:
Nice investigating... :moneybag:

PR #56, opened 2015-06-12
Refactor outcome adoption

derekprior:
* There are multiple updates involved in adopting default outcomes, so
  they should occur inside a transaction.
* Use `update_column` rather than `update_attribute`. It will raise an
  exception if the SQL update fails where update_attribute may not.
* Remove `StandardOutcome.retreive_defaults` as it's analogous to
  `.all`.


joshuaclayton:
:+1:

PR #57, opened 2015-06-12
Rename OutcomeAlignments to Alignments

derekprior:
There are no other types of alignments, so the `Outcome` prefix is
redundant. For similar reasons, the `alignment_level` field was renamed
`level`.

Data model constraints were also added.


joshuaclayton:
:rainbow:

PR #58, opened 2015-06-12
Create outcomes index

derekprior:
This page is where an admin that is managing outcomes for a given course
will see those outcomes and interact with them. It takes the place of
courses show, which was very similar.

There are currently no actions to be taken with the outcomes, but in the
future, there will be links to edit any custom outcomes.


mrgilman:
:snowman:

PR #59, opened 2015-06-12
Use jquery-turbolinks

derekprior:
The navbar menu switches to a dropdown menu on small screens. It wasn't
working on initial page load because we were bound only to the
turbolinks `page:ready` event. To handle the initial page load, we need
to also be bound to `ready`.

We could have easily bound to both events, but this is something we'd
have to remember to do any time we added JavaScript events like this.
Enter `jquery-turbolinks`. This fires the document `ready` even on
`page:load` among a few other things. You can, for the most part, just
write your jQuery as you always have.


mrgilman:
:hatched_chick:

PR #60, opened 2015-06-12
Display default outcomes before adopting

mrgilman:
https://trello.com/c/eSZ6YpCC

Add default_outcomes#show page so that users can review the standard
outcomes before adopting them for a course.


derekprior:
:+1:

PR #61, opened 2015-06-15
Adopt standard outcomes via outcomes dashboard

derekprior:
This test used to go through the course show page, which will likely be
going away as the admin section of the site is finished up today. The
workflow for adopting custom outcomes should be triggered from the
outcomes dashboard.


mrgilman:
:frog:

PR #62, opened 2015-06-15
Create alignments for standard outcomes

derekprior:
Previously, adopted standard outcomes were implicitly aligned with
standard outcome from which they were derived. This was a needless
special-casing of the standard outcomes and started to cause problem
when implementing a mix of standard and custom outcomes on a single
course.


mrgilman:
:dizzy:

PR #63, opened 2015-06-15
Introduce join table between outcomes and assessments

mrgilman:
If multiple courses within a given department have identical outcomes and those outcomes have the same assessments, we should not duplicate the assessments since the results only need to be entered once.

* Remove outcome_id from direct and indirect assessments and introduce outcome_assessments polymorphic join table
* Add department_id to direct and indirect assessments since `department` can no longer be delegated to an outcome
* Fix specs and factories to work with new schema

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150615155716) do
+ActiveRecord::Schema.define(version: 20150615190415) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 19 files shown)


db/schema.rb
```diff
@@ -70,12 +69,22 @@
...
     t.string   "minimum_requirement"
     t.integer  "target_percentage"
-    t.integer  "outcome_id"
     t.datetime "created_at",          null: false
     t.datetime "updated_at",          null: false
     t.string   "type",                null: false
   end
 
+  create_table "outcome_assessments", force: :cascade do |t|
+    t.integer  "outcome_id"
```
derekprior: null falses

derekprior:
:necktie:

PR #64, opened 2015-06-15
Make default outcomes a plural resource

derekprior:
This will pave the way for them to more easily be adopted on a
case-by-case basis. As part of this work, the outcomes dashboard now
includes a list of fully aligned courses for confirmation that you are
complete setting up outcomes for a course.


mrgilman:
This looks good to me. What do you think about `s/default_outcomes/standard_outcomes` so we don't have two terms for the same thing?

derekprior:
Yeah, I was planning that as a PR when I'm done this refactoring journey.

PR #66, opened 2015-06-15
Allow a la carte adoption of standard outcomes

derekprior:
There's likely to be scenarios where there's a desire to have a small
number of custom outcomes and backfill the rest with standard
outcomes. This change allows admins to adopt custom outcomes en masse or
a la carte to accommodate this.

Because you can adopt a standard outcome when you already have existing
outcomes, we had to remove the code that prevents this. The UI is now
responsible for displaying or hiding the ability to adopt multiple when
appropriate. We can guard against strict duplication (double clicks of
the "adopt all" button, for instance) by introducing unique constraints
at the database level.


mrgilman:
:fr:

PR #68, opened 2015-06-15
Consistently refer to Standard Outcomes

derekprior:
We were inconsistently referring to "Default Outcomes" and "Standard
Outcomes" throughout the app. "Standard Outcomes" seems more appropriate,
so this change standardizes (hah!) on that.


mrgilman:
:fries:

PR #69, opened 2015-06-16
Allow editing of outcomes

derekprior:
Outcomes and their alignments can now be edited. The desire to have all
standard outcomes listed on the outcome form leads to some interesting
controller code to ensure that each standard outcome is represented in
the alignments collection, at least in-memory.

We also wanted the alignments to be sorted, but I could not convince
Rails and `fields_for` to order the in memory collection appropriately
according to the `name` field, regardless of whether the object is
persisted or not. Ultimately, I used jQuery to handle the re-order of
the alignment fields. This works imperceptibly in my experience so I
think it's okay.

### DIFF
--- a/app/assets/javascripts/outcome_form.js
+++ b/app/assets/javascripts/outcome_form.js
@@ -1,0 +1,22 @@
+$(function(){
+  var alignmentContainer = $("fieldset.alignments");
+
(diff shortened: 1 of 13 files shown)


mrgilman:
:lollipop:

PR #70, opened 2015-06-16
Remove department and course UI

derekprior:
These views and controllers were used to support the previous
"department -> course" deep-dive UI and are no longer needed.

Currently routing the homepage to the outcomes dashboard. It will likely
make more sense to route the user depending on their access level, but
that will be easier to do once additional access roles are created and
there's an assessment landing page.


mrgilman:
:tiger:

PR #71, opened 2015-06-16
Introduce assessments dashboard

mrgilman:
https://trello.com/c/ymayP0kT

Supports workflow in which a user starts with a course and chooses the outcome they wish to add an assessment for.

Also delete some views that are no longer used and being replaced by workflows.


derekprior:
:+1:

PR #72, opened 2015-06-16
Conditional subjects index description text

derekprior:
We want to show the user different text depending on their ability to
create assessments for any department. This posed an interesting
challenge because our previous permissions checking always dealt with
whether or not the user could perform an action on a specific
department. We will need more of this type of check for other UI
changes.

Our existing code for, for instance, checking if a user is an admin of a
given department iterates over all permissions looking for one where
both the department matches and the access level is admin. In this use
case, we only care about the access level, so we pass in an object that
will *always* pass the department part of that check. This is
`AnyDepartment`.

`AnyDepartment` is explicitly marked as a private constant because I do
not want it leaking out into other parts of the app. The idea is to keep
this concept entirely contained in the policy itself.

### DIFF
--- a/app/policies/generic_policy.rb
+++ b/app/policies/generic_policy.rb
@@ -1,0 +1,23 @@
+class GenericPolicy
+  attr_reader :user
+
+  def initialize(user, _record)
+    @user = user
+  end
+
+  def create_assessments?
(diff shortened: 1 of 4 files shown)


mrgilman:
:octopus:

PR #73, opened 2015-06-16
Update Gems

derekprior:
* Updated to recently-released versions of Rack, Rails, jquery-rails, and
  web-console to address security issues disclosed today
* Updated to latest version of RSpec
* Removed unused selenium-webdriver gem


mrgilman:
:surfer:

PR #74, opened 2015-06-16
Introduce route/controller namespacing for assessment workflow

mrgilman:
Namespacing routes allows us to use the same RESTful resources in multiple workflows.


derekprior:
:+1:

PR #75, opened 2015-06-17
Separate direct and indirect assessment Pundit policies

mrgilman:
Since direct assessments no longer belong to one outcome, we cannot base
authorization on the associated outcome's department. This change bases
authorization for creating new direct assessments on the user's
permission level and not their departments.

### DIFF
--- a/app/policies/direct_assessment_policy.rb
+++ b/app/policies/direct_assessment_policy.rb
@@ -1,0 +1,17 @@
+class DirectAssessmentPolicy < ApplicationPolicy
+  def new?
+    GenericPolicy.new(user, record).create_assessments?
+  end
+
+  def create?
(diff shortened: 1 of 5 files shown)


app/policies/direct_assessment_policy.rb
```diff
@@ -0,0 +1,13 @@
+class DirectAssessmentPolicy < ApplicationPolicy
+  def create?
+    GenericPolicy.new(user, record).create_assessments?
```
derekprior: At the point of creation you *will* know enough information to determine the applicable department, correct? So we'd want to go based on that, right? What about manually using the generic policy in the new action itself and then using the more specific `DirectAsessmentPolicy` on create?

derekprior:
:cake:

PR #77, opened 2015-06-17
Enable creation of direct assessment starting with subject

mrgilman:
https://trello.com/c/ymayP0kT

Instead of having a user choose if they want to create an assessment "by
course" or "by subject", we now take users through two different
workflows when adding direct and indirect assessments.

Indirect assessment creation is similar to the previous workflow; users
choose a course and an outcome and add an assessment.

When users create new direct assessments, they are prompted first to
choose a subject from a dropdown, then are shown all outcomes for the
departments they have access to with check boxes, so they can pick the
ones they want to associate with the new assessment. We used a form
object with an `outcomes_ids` attribute to handle the creation of those
associations.

Since direct assessments have a `department_id` column and we don't know
the department until the assessment is created, we validate that the
outcomes chosen belong to the same department, and set the department of
the new assessment to that department.

### DIFF
--- a/app/controllers/direct_assessments_controller.rb
+++ b/app/controllers/direct_assessments_controller.rb
@@ -1,36 +1,39 @@
 class DirectAssessmentsController < ApplicationController
   def new
-    @outcome = Outcome.find(params[:outcome_id])
-    @assessment = @outcome.direct_assessments.build
-    authorize(@outcome)
+    @assessment = DirectAssessment.new
+    @courses = scoped_courses
+    authorize(@assessment)
   end
 
   def create
-    @outcome = Outcome.find(params[:outcome_id])
-    @assessment = @outcome.direct_assessments.build(direct_assessment_params.merge(department_id: @outcome.department.id))
-    authorize(@outcome)
+    @assessment = DirectAssessment.new(direct_assessment_params)
+    @courses = scoped_courses
+    authorize(@assessment)
 
-    if @outcome.save
-      redirect_to outcome_path(@outcome), success: t(".success")
+    if @assessment.save
+      redirect_to direct_assessment_path(@assessment.id), success: t(".success")
     else
+      flash.now[:error] = @assessment.errors.full_messages.join("\n")
       render :new
     end
   end
 
   def edit
     @assessment = DirectAssessment.find(params[:id])
+    @courses = @assessment.department.courses.with_outcomes
     authorize(@assessment)
   end
 
   def update
     @assessment = DirectAssessment.find(params[:id])
+    @courses = @assessment.department.courses.with_outcomes
     authorize(@assessment)
 
-    @assessment.assign_attributes(direct_assessment_params)
-    if @assessment.save
-      redirect_to outcome_path(@assessment.outcomes.first), success: t(".success")
+    if @assessment.update(direct_assessment_params)
+      redirect_to direct_assessment_path(@assessment.id), success: t(".success")
     else
-      render :edit, success: t(".success")
+      flash.now[:error] = @assessment.errors.full_messages.join("\n")
+      render :edit
     end
   end
 
@@ -44,14 +47,18 @@ class DirectAssessmentsController < ApplicationController
   def direct_assessment_params
     params.require(:direct_assessment).permit(
       :actual_percentage,
-      :department_id,
       :description,
       :minimum_requirement,
       :name,
       :problem_description,
       :semester,
       :subject_id,
-      :target_percentage
+      :target_percentage,
+      outcome_ids: []
     )
+  end
+
+  def scoped_courses
+    policy_scope(Course).with_outcomes.includes(:outcomes)
   end
 end
--- a/app/models/direct_assessment.rb
+++ b/app/models/direct_assessment.rb
@@ -2,15 +2,35 @@ class DirectAssessment < ActiveRecord::Base
   has_many :outcome_assessments, as: :assessment
   has_many :outcomes, through: :outcome_assessments
 
-  belongs_to :department
   belongs_to :subject
   has_many :results, as: :assessment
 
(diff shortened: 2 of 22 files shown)


derekprior:
I worked on this with you, so I'm maybe not the best review, but I think it's :+1:. I wish that form object wasn't such a pain, but such is life.

app/controllers/direct_assessments_controller.rb
```diff
@@ -1,36 +1,40 @@
 class DirectAssessmentsController < ApplicationController
   def new
-    @outcome = Outcome.find(params[:outcome_id])
-    @assessment = @outcome.direct_assessments.build
-    authorize(@outcome)
+    @assessment = DirectAssessmentForm.new(DirectAssessment.new, [])
```
joshuaclayton: Can this array actually also be `outcome_ids`?

app/models/direct_assessment_form.rb
```diff
@@ -0,0 +1,72 @@
...
+    DirectAssessment.model_name
+  end
+
+  def self.from(assessment)
+    new(assessment, assessment.outcome_assessments.pluck(:outcome_id))
+  end
+
+  def initialize(direct_assessment, outcome_ids)
+    @direct_assessment = direct_assessment
+    @outcome_ids = Array.wrap(outcome_ids).reject(&:blank?)
```
joshuaclayton: Since `#outcome_ids` already does this, thoughts on not modifying what's passed into the constructor?

Alternatively, `#outcome_ids` could change to an attr_reader and we modify in the constructor, but I'm guessing we're assigning to it in `#update`

app/controllers/direct_assessments_controller.rb
mrgilman: I think since the params are nested we'd need to do something like `params.fetch(:direct_assessment, {}).fetch(:outcome_ids, []), so I think passing the empty array is better. We could also have it as a default in the initializer.

app/models/direct_assessment_form.rb
mrgilman: This is a good call -- both `attr_accessor`s in the form object can actually be readers.

derekprior:
Well, this was fun. In the end, we realized we could use the `outcome_ids` method provided by the `has_many :through` association on `DirectAssessment`. There's some ugliness to do with `before_save` in order to set the department, but we may be able to avoid that...

app/models/direct_assessment.rb
```diff
@@ -7,4 +7,26 @@ class DirectAssessment < ActiveRecord::Base
   has_many :results, as: :assessment
 
   validates :subject, presence: true
+  validate :ensure_single_department
+  before_save :set_department
```
derekprior: I feel like we had this discussion just a day or two ago... but with our validation now ensuring all associated outcomes are for the same department, do we still need assessments to have a direct reference to department? Or could we define `department` as `outcomes.first.department`?

Do we ever search for assessments by department?

app/controllers/direct_assessments_controller.rb
joshuaclayton: @mrgilman makes sense!

app/models/direct_assessment.rb
mrgilman: I think you're right and we can drop that column.

derekprior: :+1:

app/views/direct_assessments/_form.html.erb
```diff
@@ -1,58 +1,22 @@
...
+  <% end %>
+
+  <br>
+
+  <%= f.input :name, label: "Assignment name", placeholder: "e.g. Problem Set 1" %>
+  <%= f.input :description, label: "Assignment description", placeholder: "e.g. 'Integration'" %>
+  <%= f.input :problem_description, placeholder: "e.g. 'Question 3, Integration by parts'" %>
+  <%= f.input :minimum_requirement, label: "Minimum grade", placeholder: "e.g. '7 points out of 10'" %>
+  <%= f.input :target_percentage, placeholder: "Whole number, e.g. '80' for '80%'" %>
+  <%= f.submit "Submit" %>
```
derekprior: can you make this:

```erb
<div class="actions">
  <%= f.submit "Submit" %>
  <%= cancel_button %>
</div>
```

So it's consistent with the cancel button stuff I just merged?

app/views/direct_assessments/show.html.erb
```diff
@@ -1,25 +1,43 @@
...
-  <tr>
-    <th>Name</th>
-    <th>Description</th>
-    <th>Problem Description</th>
-    <th>Year</th>
-    <th>Semester</th>
-    <th>Percentage</th>
-  </tr>
-  <% @assessment.results.each do |result| %>
+<% if @assessment.results.any? %>
```
derekprior: I think that this should be `present?`. We're going to iterate it if there are any, which means we'd need to go back to the database to get all the records. What do you think?

spec/spec_helper.rb
```diff
@@ -10,4 +10,5 @@
   end
 
   config.order = :random
+  config.example_status_persistence_file_path = "spec/examples.txt"
```
derekprior: Would this work fine if we put this in `tmp/spec_examples.txt` ?

app/views/direct_assessments/_form.html.erb
mrgilman: :+1:

app/views/direct_assessments/show.html.erb
mrgilman: Sure, makes sense.

spec/spec_helper.rb
mrgilman: Yup, it should. I'll change it.

PR #79, opened 2015-06-17
Result form improvement

derekprior:
* Make it under the "Record Data" Tab
* Add some field sets to group the fields
* Add validations to display required fields
* Enable HTML5 validations
* Add cancel button


mrgilman:
:bowtie:

PR #80, opened 2015-06-18
Add cancel buttons to forms

derekprior:
The cancel buttons have a relatively sane default `href` but are
progressively enhanced with JavaScript to be equivalent to pressing the
back button in your browser, which is more likely to be accurate in all
cases.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -51,18 +51,13 @@ en:
       description: Description
       heading: Outcomes for %{name}
       label: Label
-  departments:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -40,7 +40,7 @@ en:
         assess: Assess
         details: Details
   assessments_dashboard:
-    show:
+    show:
```
joshuaclayton: trailing whitespace

PR #82, opened 2015-06-18
Improve direct assessment result workflow

derekprior:
* Add tabs to all pages
* Internationalize text
* Hide actions from users not authorized for them
* Improve styling

### DIFF
--- a/app/models/permission_set.rb
+++ b/app/models/permission_set.rb
@@ -39,7 +39,4 @@ class PermissionSet
   def admin?(department)
     any? { |permission| permission.admin?(department) }
   end
-
-  # temporary - this method is in a different PR I will merge first
-  alias_method :manage_results?, :admin?
 end
--- a/spec/features/ta_creates_a_result_spec.rb
+++ b/spec/features/ta_creates_a_result_spec.rb
@@ -1,11 +1,11 @@
 require "rails_helper"
 
-feature "User creates a result" do
-  scenario "and it is associated with a direct assessment" do
+feature "TA creates a result" do
+  scenario "for a direct assessment" do
     assessment = create(:direct_assessment)
     department = assessment.department
     permitted_subject = assessment.subject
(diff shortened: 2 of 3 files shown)


app/views/subjects/show.html.erb
```diff
@@ -1,23 +1,28 @@
-<h1>Direct Assessments for <%= @subject %></h1>
-<% if @subject.direct_assessments.any? %>
+<% tab(TabHelper::DATA_ENTRY) %>
+
+<h1><%= @subject %></h1>
+
+<% if @subject.direct_assessments.present? %>
```
joshuaclayton: `#any?` instead of `#present?`?

spec/features/user_creates_a_result_spec.rb
```diff
@@ -77,20 +60,4 @@
     expect(page).to have_content "Annual survey"
     expect(page).to have_content "92"
   end
-
-  scenario "and it defaults to the same name, description as the assessment" do
```
joshuaclayton: are these tests duplicated somewhere else?

app/views/subjects/show.html.erb
derekprior: That will trigger an exists query, which if there are any, will end up doing another round trip when I iterate the collection below. `present?` will trigger one query with either all of my data or an empty list.

spec/features/user_creates_a_result_spec.rb
derekprior: I thought they were, but looking at it, they are not. I can do it more directly with a controller spec, though.

mrgilman: I think we need these tests to ensure that the results form is pre-filled with the default attributes from the assessment.

derekprior: I added these expectations back to the existing tests.

PR #83, opened 2015-06-18
Add validations to outcomes form

derekprior:
This adds missing database constraints and indexes on outcomes and then
backs them up with validations in the form. This leads to a better user
experience - they have a visual indication of required fields and get
nice errors rather than server errors when things fail.

As part of this change, I configured simple_form to not display error
text for each individual field, instead opting for a summary at the top
of the form. This isn't ideal; Inline error messages would be best with
the summary being reserved for things that relate to more than one
field. However, this was simpler to style in a reasonable manner for
now.

### DIFF
--- a/app/helpers/errors_helper.rb
+++ b/app/helpers/errors_helper.rb
@@ -1,0 +1,9 @@
+module ErrorsHelper
+  def render_errors(form)
+    errors = form.object.errors
+
+    if errors.any?
+      render "errors", errors: errors.full_messages
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


app/helpers/errors_helper.rb
```diff
@@ -0,0 +1,7 @@
+module ErrorsHelper
+  def render_errors(form)
+    if form.object.errors.any?
```
joshuaclayton: thoughts on assigning `form.object.errors` to a local variable to use in the conditional and render?

joshuaclayton:
one comment but LGTM

PR #84, opened 2015-06-18
Add auditing via paper trail

derekprior:
There was a request for user tracking so we could track who created and
updated user-created records (non-seed data). It turns out that it's far
simpler to use paper_trail to create a full audit trail for the given
models.


joshuaclayton:
:+1:

PR #86, opened 2015-06-18
Fix form actions styling

derekprior:
Using the `.actions` class for the form buttons container was
overlapping in some places with out use of the `.actions` class on table
contents.

Resolved quickly by using `.form-actions` for the form buttons.


mrgilman:
:

PR #88, opened 2015-06-18
Scope courses to those with permissions in direct assessment form

mrgilman:
Modify `PermissionSet#department_slugs` so that it takes an optional
access_levels argument instead of returning slugs for which the user has
any permission level, and use this new functionality to scope the
courses that are rendered in the direct_assessments form.


derekprior:
:+1:

PR #89, opened 2015-06-18
Add help text to outcomes dashboard

derekprior:
Try to let people know what each of these sections represents since the
users may not be super familiar with ABET.


derekprior:
There's a lot of text here (particularly in the first case) but I think it's okay for now. It may make sense to put a "?" icon to the right of the header that would toggle or show the additional text for each table but I think this is a good start.

abet_outcomes_assessment

derekprior:
This isn't a super interesting change (mostly text) so I'm going to merge this in without further review.

PR #90, opened 2015-06-18
Add missing constraints and validations

derekprior:
There were some missing null constraints, foreign keys, and indexes on
the database level. With those added, I backfilled direct assessment and
indirect assessment with user-facing validations on the required fields.


mcmire:
Seems pretty straightforward to me :+1:

PR #93, opened 2015-06-18
Display list of assessments by course

mrgilman:
https://trello.com/c/y65cLV2q

From the "Manage Outcomes" page, have the "Edit Assessments" link point
to an assessments index page which displays all assessments associated
with a course by outcome. From here, users will be able to update
assessments and the outcomes they are associated with.


derekprior:
looks good. I'll merge it.

PR #95, opened 2015-06-19
Improve check box list display creating assessment

derekprior:
Previously, long outcomes would wrap around to have text under the
check box which broke up the list of check boxes down the left rail. With
these changes the check boxes all line up down the left rail and long
labels do not wrap under the check boxes.


derekprior:
abet_outcomes_assessment

mrgilman:
:8ball:

PR #96, opened 2015-06-19
Namespace outcome management actions

derekprior:
We're finding it difficult to discern which resources and views are used
in which work flows. We're also occasionally using the wrong resource
name in routes because a different workflow is already using the
resource name we want to use.

Moving to namespaces for the various workflows can mitigate this. It's a
bit of a pain because of the long route helpers and the fact that you
may now have two files called "outcomes_controller.rb", but I think the
trade off will be worth it here.


mrgilman:
:baseball:

PR #97, opened 2015-06-19
Change success flash color

derekprior:
The green flashes with the red navigation bar and red headers had a very
Christmas feel to them, but it's June. Let's use the notice (blue) color
instead.


mrgilman:
:green_book: :blue_book:

mborsare:
:+1:

PR #99, opened 2015-06-19
Enable deletion of results from direct and indirect assessments

mrgilman:
https://trello.com/c/IQrfFM5C

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -33,6 +33,14 @@ class ResultsController < ApplicationController
     end
   end
 
+  def destroy
+    result = Result.find(params[:id])
+    authorize(result)
+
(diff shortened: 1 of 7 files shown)


app/controllers/results_controller.rb
```diff
@@ -33,6 +33,14 @@ def update
     end
   end
 
+  def destroy
+    result = Result.find(params[:id])
+    authorize(result)
+
+    result.destroy
```
derekprior: How about `destroy!` since we're not checking the return value?

derekprior:
1 comment, otherwsie :+1:

app/controllers/results_controller.rb
mrgilman: ok!

PR #101, opened 2015-06-19
Allow users with view access to see assessments

mrgilman:
Hide assessment create/update links from users with read-only access


derekprior:
:hamster:

PR #102, opened 2015-06-19
Create manage results namespace

derekprior:
Each of the workflows is now segregated into its own namespace to make
it clear what roles each workflow has and when we are crossing workflow
boundaries. The named route helpers are really terrible, but it's a good
tradeoff.


mrgilman:
:feet:

PR #103, opened 2015-06-19
Display number of assessments for outcomes

mrgilman:
Add counter cache of `assessments_count` to outcomes, and add `dependent: :destroy` to the join relationship so that the cache correctly decreases when an assessment is deleted.


derekprior:
:+1:

PR #117, opened 2015-12-04
Update redirects used after managing assessments

derekprior:
This previously brought the user to a screen to create or view results
for this assessment. However, it's far more likely that you will want to
stay in the context of managing assessments.

For creating direct assessments, we now redirect back to the form with
the previous subject pre-selected. The flash message has links to return
to the listing of assessments for the courses that were selected in the
previous assessment. This gives users the ability to enter many new
assessments for a subject in a less painful manner.

Other redirects were also updated to bring the user to more logical
assessment locations.


derekprior:
abet_outcomes_assessment

PR #123, opened 2015-12-09
Add assessment report

derekprior:
This adds an HTML view for a slimmed down report of assessment data for
a given course and makes the full report downloaded via a link to CSV.
The HTML version is purely to give a sense of the type of data present
in the report (and if there's any new data since they last looked at
it). The CSV contains more data, allowing them to do what they would
like in excel.

None of this UI is linked in to the main site at this point because I
wasn't quite certain where in the "Manage Assessments" flow to put it...
or even if it should go there. To see this in action, you'll need to go
to `/reports/course/:course_id/assessment_report`


derekprior:
abet_outcomes_assessment

derekprior:
If you have feedback on what fields should be in the HTML version, let me know. I was trying to give a sense of "there's data here and here's some that might be useful" before triggering a download.

PR #124, opened 2015-12-09
Track number of active assessments for an outcome

derekprior:
We were previously tracking the number of assessments associated to an
outcome with a counter cache on `OutcomeAssessment`. With the introduction
of the notion of archived assessments, this count no longer reflects the
number of active assessments.

Getting counter caches to obey conditions is tricky. Since the condition we
care about is binary it's not too bad to do it manually. We keep
`assessments_count` counting the overall number of assessments and add a
service object (AssessmentArchivist) that is responsible for archiving an
unarchiving that tracks a separate count of the number of archived
outcomes. Then it's simple math to get the number of active outcomes.

### DIFF
--- a/app/controllers/manage_assessments/archives_controller.rb
+++ b/app/controllers/manage_assessments/archives_controller.rb
@@ -1,10 +1,21 @@
 module ManageAssessments
   class ArchivesController < ApplicationController
     def create
-      assessment = DirectAssessment.find(params[:direct_assessment_id])
       authorize(assessment, :update?)
-      assessment.update_attributes(archived: true)
+      assessment.archive
       redirect_to_back_or_default root_path, success: t(".success")
+    end
+
+    def destroy
+      authorize(assessment, :update?)
+      assessment.unarchive
+      redirect_to_back_or_default root_path, success: t(".success")
+    end
+
+    private
+
+    def assessment
+      @_assessment ||= DirectAssessment.find(params[:direct_assessment_id])
     end
   end
 end
--- a/app/controllers/manage_assessments/direct_assessments_controller.rb
+++ b/app/controllers/manage_assessments/direct_assessments_controller.rb
@@ -59,7 +59,6 @@ class ManageAssessments::DirectAssessmentsController < ApplicationController
 
   def direct_assessment_params
     params.require(:direct_assessment).permit(
-      :archived,
       :description,
       :minimum_requirement,
       :name,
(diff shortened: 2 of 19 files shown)


derekprior:
This isn't quite ready to be merged. I thought of one additional gotcha:

1. assessment is created covering outcome a and b ( each outcome has `assessments_count: 1`, `archived_assessments_count: 0`)
2. assessment is archived (each outcome has `assessments_count: 1`, `archived_assessments_count: 1`)
3. assessment is edited, and outcome a is removed (outcome a has `assessments_count: 0`, `archived_assessments_count: 1` for a total active of `-1`, which is wrong).

We could disallow editing via the UI for archived assessments, but I'm not sure that's okay to do. @jdaries what do you think?

@joshuaclayton Would love some input here if you have any. Any other ideas?

derekprior:
Another fix would be a before_destroy callback on the model which would decrement the associated `outcomes.archived_assessments_count` if the assessment was archived.

Am I missing something more obvious? Would a query be better at this point?

jdaries:
Yes, fine to lock editing after archived. Similar to how Mac OS treats files in the Trash--you want to mess with it? Then it doesn't belong in the trash. A user could unarchive and then edit and rear hive of for whatever silly reason they need to do that.

On Dec 9, 2015, at 6:06 PM, Derek Prior <notifications@github.com<mailto:notifications@github.com>> wrote:


This isn't quite ready to be merged. I thought of one additional gotcha:

  1.  assessment is created covering outcome a and b ( each outcome has assessments_count: 1, archived_assessments_count: 0)
  2.  assessment is archived (each outcome has assessments_count: 1, archived_assessments_count: 1)
  3.  assessment is edited, and outcome a is removed (outcome a has assessments_count: 0, archived_assessments_count: 1 for a total active of -1, which is wrong).

We could disallow editing via the UI for archived assessments, but I'm not sure that's okay to do. @jdaries<https://github.com/jdaries> what do you think?

@joshuaclayton<https://github.com/joshuaclayton> Would love some input here if you have any. Any other ideas?

derekprior:
Ok, I'll likely tackle that separately

joshuaclayton:
@derekprior could we do a `[0, (assessments - archived_assessments)].max` to calculate count, ensuring the floor is 0?

joshuaclayton:
@derekprior actually, does https://github.com/MIT-IR/abet/pull/124/files#diff-ef06a303142e4ab5f509d829e66d9741R27 do what I'd suggested re: max?

derekprior:
Yeah. I was thinking of doing that and will do so, but it only addresses that specific off-by-one instance.

The `GREATEST` bit ensures that `archived_assessments` can never be negative. Not that calculated active assessments won't be.

PR #127, opened 2015-12-15
Add OutcomeWithMetadata model that has live stats

derekprior:
My previous attempt to use something akin to a counter cache worked okay
for counting archived assessments and thus arriving at the number of
active assessments. The next step was to get a count of those
active assessments that also have at least one result. Doing this was
going to be difficult without introducing extraneous queries or setting
up many active record associations exclusively for this purpose.

Instead, this change introduces and `OutcomeWithMetadata` model that is
backed by a SQL view of outcomes along with the various statistics we
need. The model inherits from `Outcome` and thus has all usual `Outcome`
methods and associations. We can now use `OutcomeWithMetadata.find(id)`
when we want to get the accompanying statistics.

Overall, I feel like keeping this data alongside an outcome by using SQL
is less confusing that encapsulating all interaction in service objects
and making the application maintain counters. The SQL stats should never
go out of date, regardless of how the API in the application is used.


derekprior:
@joshuaclayton Decided to reverse field a bit here with the next requirement. I'm curious what you think of this approach. I ran it by Sean as well and his advice was that this or a series of database triggers for managing coutners (rather than app code) would likely be the best approach given the number of these conditional counters we want.

PR #128, opened 2015-12-15
Fixes to progress bar

alexandra-aas:
`0.to_f / 0.to_f` returns `NaN`, so the progress bar for `0 / 0` was showing as 100%. Also wanted to limit the width of the progress bar.

Before:
screen shot 2015-12-15 at 4 14 48 pm

After:
screen shot 2015-12-15 at 4 14 59 pm


derekprior:
:+1:

PR #129, opened 2015-12-16
Streamline Manage Assessments flow

alexandra-aas:
- Introduced a modal for creating a new assessment. This makes landing page for "Managing Assessments" simpler. The same button is also carried into the "Managing Assessments for [Course]" view, which helps to keep that action consistent and eliminates the need for the user to be redirected back to the landing page from that view in order to create a new assessment.
new_assessment_modal
screencapture-localhost-3000-manage_assessments-1450300090635

- Restructured the "Manage Assessments for [Courses]" view to have two tables: direct and indirect. Then within those the assessments are separated by outcome. The outcome description is no longer present, but you can hover over the title row (Outcome A (1 Assessment)) and the description will show up. If that needs to be more prevalent than I can make a fix for that. Overall it takes up less room and is easier to read while representing the same information as before. Since Derek is working on the gradebook and probably shouldn't fuss with filtering the page right now, I think this is a good first step towards a cleaner page.

- Added "Edit" and "Archive" actions to the assessment. Also moved the "Add Results" button above the results table. When there are more results the button won't get buried farther down the page.
screen shot 2015-12-16 at 2 29 26 pm

### DIFF
--- a/app/views/manage_assessments/dashboard/show.html.erb
+++ b/app/views/manage_assessments/dashboard/show.html.erb
@@ -1,34 +1,47 @@
 <% tab(TabHelper::ASSESSMENTS) %>
-<% if policy(:generic).create_assessments? %>
-  <h2>Create New Assessment</h2>
-  <div class="assessments-dashboard-options">
-    <div class="assessments-dashboard-option">
-      <h6><%= t(".by") %></h6>
-      <h1><%= t(".subject") %></h1>
-      <%= link_to t(".get_started"),
-        new_manage_assessments_direct_assessment_path,
-        class: "button",
-        data: { role: "start-direct-assessment" }
-      %>
-    </div>
 
-    <div class="assessments-dashboard-option">
-      <h6><%= t(".by") %></h6>
-      <h1><%= t(".course") %></h1>
-      <%= link_to t(".get_started"),
-        manage_assessments_courses_path,
-        class: "button",
-        data: { role: "start-indirect-assessment" }
-      %>
-    </div>
-  </div>
-<% end %>
+<div class="headline-with-action">
+  <h1><%= t(".title") %></h1>
 
-<h2>Manage Existing Assessments</h2>
+  <% if policy(:generic).create_assessments? %>
+    <%= link_to t(".create_new_assessment"), javascript_void,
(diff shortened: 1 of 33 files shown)


app/views/manage_assessments/dashboard/show.html.erb
```diff
@@ -1,34 +1,35 @@
...
+        <th><%= t(".actions") %></th>
+      </tr>
+    </thead>
+
+    <tbody>
+      <% @courses.each do |course| %>
+        <tr>
+          <td><strong><%= course.number %></strong></td>
+          <td><%= course.name %></td>
+          <td><%= render "progress_bar",
```
alexandra-aas: @derekprior Is there an easy way to get assessment count for a course?

derekprior: It can be calculated somewhat simply.

derekprior: I can add it as another commit

derekprior:
Good to merge!

alexandra-aas:
@derekprior I was just waiting for @jdaries to give the design okay since it's short of what we originally discussed.

jdaries:
I can look first thing in the morning.

On Dec 16, 2015, at 6:19 PM, Ali <notifications@github.com<mailto:notifications@github.com>> wrote:


@derekprior<https://github.com/derekprior> I was just waiting for @jdaries<https://github.com/jdaries> to give the design okay since it's short of what we originally discussed.

jdaries:
So there are two problems that I came across with this PR, and the new Manage Assessments interface doesn't check for the existence of outcomes. If a course does not have outcomes aligned yet, there should not be the option to make a new assessment. 

Also, if there are multiple courses available for a user, what does the big blue "make assessment" button on the right use? If it is defaulting to the first one, there's no indication of which course has been chosen in the create assessment form.

alexandra-aas:
@jdaries Just pushed a commit that should fix the issues we discussed.

screen shot 2015-12-17 at 12 40 06 pm

alexandra-aas:
Went with "Create New Assessment" even though it's a little weird.

screen shot 2015-12-17 at 1 08 02 pm

PR #131, opened 2015-12-17
Allow assessment archiving from assessments list

derekprior:
Previously this functionality was only accessible via the results
screen.

### DIFF
--- a/app/views/manage_assessments/assessments/_direct_assessments.html.erb
+++ b/app/views/manage_assessments/assessments/_direct_assessments.html.erb
@@ -1,5 +1,5 @@
 <tr>
-  <td colspan="7" class="table-section-header" title="<%= outcome.description %>">
+  <td colspan="3" class="table-section-header" title="<%= outcome.description %>">
     <%= t(".outcome_table_section_header",
       outcome: outcome.name,
       count: pluralize(outcome.direct_assessments.count, t(".assessment"))) %>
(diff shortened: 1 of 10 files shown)


derekprior:
abet_outcomes_assessment

@alexandra-aas I'm getting wrapping in the table now. Any ideas for an easy fix? I tried "whitespace: nowrap" on the actions div and that fixes that column but the others wrap instead. Do you have any ideas?

Wonder if maybe we can live without the target percentage and minimum grade columns right there? What do you think @jdaries ?

alexandra-aas:
@derekprior I'll take a look. We could also just switch back to not inlining them and having them stacked again. My intention was to make the rows less tall.

jdaries:
@derekprior i would be okay with eliminating those two columns

app/views/manage_assessments/assessments/_direct_assessments.html.erb
```diff
@@ -13,11 +13,17 @@
     <td><%= assessment.minimum_requirement %></td>
     <td><%= assessment.target_percentage %></td>
     <td class="inline-action-links">
```
alexandra-aas: You can do: `<td class="inline-action-links" nowrap>`.

alexandra-aas:
@derekprior The above fixes the issue, but I would also be for eliminating those two rows because I don't think that information is crucial at that level.

One other thing. I added an archive button to `manage_results/subjects/show.html.erb` in my last PR. Do you want to double check that I have the right code there considering these changes.

derekprior:
I'll fix this up and take a look at that, yes. Thanks

PR #133, opened 2015-12-21
Allow users to undo an archive action

derekprior:
Very similar to GMail, allow users to immediately undo an action they
may have done in error.


derekprior:
abet_outcomes_assessment

@alexandra-aas It doesn't look like you had mocked up in the trello card. Is this fine or do you want to try to re-style the flashes?

We can use a different key as well, if you want it to be something other than a success flash.

alexandra-aas:
@derekprior I think that works better because it stays in line with the rest of the app. :+1: From my end.

PR #134, opened 2015-12-22
Rough in Gradebook Histogram

derekprior:
We can assist users in reporting results on a direct assessment by
allowing them to view the gradebook grade distribution histogram from
the gradebook service. This will be particularly helpful to users who do
not have access to the gradebook (because it provides personal student
information) but can have access to aggregate grade data for the
purposes of completing ABET assessments.

This change works off of local fixtures rather than integrating with the
actual gradebook service. That will be in the next change.


app/policies/subject_policy.rb
```diff
@@ -9,9 +9,7 @@ def manage_assessments?
 
   class Scope < Scope
     def resolve
-      Subject.
-        with_direct_assessments.
-        where(department_number: user.departments.pluck(:number))
+      Subject.where(department_number: user.departments.pluck(:number))
```
jdaries: Will opening this scope up mean that when you go to add results, all subjects will display, not just those with direct assessments?

derekprior: It should not. I made a corresponding change adding this scope back in to where we need it - the subjects controller.

PR #136, opened 2015-12-23
Add result activity feed

derekprior:
Users wanted to quickly be able to see if results were being added for
the departments they have access to.

The result activity feed is built off of data that PaperTrail already
has for results. It tracks every create, update, and delete so we don't
have to. I created an `Activity` model that has some additional smarts
about what we want to get out PaperTrail data (the associated user, the
associated, non-polymorphic result).

To accommodate authorization in a performant way, I denormalized the
department a result is associated to. Previously the result department
came from `assessment -> outcomes -> courses -> departments`. Due to the
polymorphic assessment relationship, we weren't going to be able to
efficiently (read: In SQL) limit the results to only those in
departments the user can access. By caching the department at save
time, we can quickly get at this data.

One downside of this is that we can't easily show activities for deleted
results. The result is no longer in the results table, so we can't tell
what department it belongs to. We may be able to rehydrate this data
from the PaperTrail version, but doing so would be non-trivial. Earlier
Jon had mentioned that maintaing deletes was not a priority, so I went
with this for now.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -38,6 +38,16 @@ en:
       other_assessment: Other Assessment
       participation: Program Participation
       survey: Survey
+  activities:
+    activity:
+      event:
+        create_html: Results created for %{assessment}
+        update_html: Results updated for %{assessment}
+    index:
+      activity_count:
+        one: 1 Change
+        other: "%{count} Changes"
+      heading: Activity Feed
   application:
     navigation:
       assessments: Assessments
@@ -50,6 +60,7 @@ en:
       outcomes: Outcomes
       recording_data: Recording Data
     tabs:
+      activity_feed: Activity Feed
       assessments: Manage Assessments
       data: Record Data
       outcomes: Manage Outcomes
@@ -63,6 +74,8 @@ en:
   helpers:
     cancel_button: Cancel
     jon_daries_html: Jon Daries in Institutional Research at %{email}
+    today: Today
+    yesterday: Yesterday
   layouts:
     documentation_side_nav:
       assessments: Assessments
--- a/app/assets/images/icon_create.svg
+++ b/app/assets/images/icon_create.svg
@@ -1,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
+  <path d="M9 7V5H7v2H5v2h2v2h2V9h2V7H9zm-1 9c4.418 0 8-3.582 8-8s-3.582-8-8-8-8 3.582-8 8 3.582 8 8 8z" fill="#417505" fill-rule="evenodd"/>
+</svg>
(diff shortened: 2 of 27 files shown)


derekprior:
@alexandra-aas I didn't really bother trying to get this to look right and the markup is probably not at all what you want, but you should have enough to get going on the design here again.

A few notes on differences with the mockup:

1. We don't have users actual names. We only have their usernames (and emails).
2. I didn't display the associated subject. This is only applicable for DirectAssessments. I can add this, I just haven't yet.
3. I didn't do the aggregate stats at the top of the page. That's different enough to be its own change and I'd like to have Jon prioritized that separately.
4. As mentioned in the PR, there are no delete activities.

alexandra-aas:
screen shot 2016-01-04 at 1 36 30 pm

config/locales/en.yml
```diff
@@ -38,6 +38,17 @@ en:
       other_assessment: Other Assessment
       participation: Program Participation
       survey: Survey
+  activities:
+    activity:
+      event:
+        create_html: Results created for %{assessment}
+        update_html: Results updated for %{assessment}
+        destroy_html: Results deleted for %{assessment}
```
alexandra-aas: @derekprior This can be deleted, correct?

derekprior: Yes.

jdaries:
@alexandra-aas Can we have it be firstname and lastname of the person? Just first name won't be specific enough. Please excuse me if i'm taking a wireframe too literally.

alexandra-aas:
@jdaries Actually, it will show your mit username, so it will be like below. I don't think we capture first and last, so that's as unique as we can get. Is that correct @derekprior?

screen shot 2016-01-04 at 1 47 01 pm

derekprior:
That is correct. We don't have first and last name currently, but I just looked up the docs for the touchstone authentication and if our server is configured properly, we should actually get a "display name" in the request that we could use.

We didn't collect that for previously existing users so we'd have to come up with a way to back-fill it as existing users return to the site and fall back to username where its not present. I could add this as another change if there's time after the histogram stuff.

@jdaries do you want to create an prioritize a card for that?

derekprior:
The docs I referenced: https://wikis.mit.edu/confluence/display/TOUCHSTONE/Touchstone-attributes-core

jdaries:
Nope, username is good! Just didn't want it to be first name only.

jaclynperrone:
@alexandra-aas Looking good! Just a few rhetorical questions/things to consider. (Also not quite sure what to call some things, so I'm referring to the blue links as "assignments", and the gray text underneath as "departments")

* What is the most important information on this page? Some elements (author, assignment) compete for my attention.
* Can you consolidate repeated information? ie. department or assignment name?
* Perhaps you can integrate the author name into the activity (ie. "Jon edited results for ..."), since they are inherently linked.
* Nice use of icons!
* Is it important to see the # count of changes? Especially if you can read what the changes are directly below?

nice work!

alexandra-aas:
screencapture-localhost-3000-activities-1452022483842

derekprior:
@alexandra-aas This looks good. One small caveat is that not all results have subjects, so I'll need to add some conditional logic there. Mind if I take this over and merge it in? Are you set on the design front?

alexandra-aas:
@derekprior Sounds good. Thanks!

PR #137, opened 2015-12-23
Allow users to filter assessments by outcome

derekprior:
This adds a form to the assessments index that lets users filter to a
specific outcome.


derekprior:
@alexandra-aas Here's the development work for filtering by an outcome on the assessments page. It looks terrible, but it works. Have at it when you are ready.

Search will be another change.

abet_outcomes_assessment

alexandra-aas:
@derekprior Thanks. Thoughts:

- Can we do direct assessments for course and indirect assessments for course? I want to avoid looping through outcomes to show assessments because then they can show up multiple times. Not sure how much this complicates the filtering.
- That would also be ideal so we can have the count of direct and indirect and also show a "No results for filter" message in the table if nothing matches.

alexandra-aas:
Design Changes:
- Assessments are split into two tables: Direct & Indirect
- Each Assessment is only listed once, which significantly reduces page length and should eliminate confusion
- Outcome filter allows the user to see which assessments for the course cover which outcomes

assessments_for_course

alexandra-aas:
@derekprior Two issues:

1) The count on "Manage Assessments for Your Courses" doesn't match the count when you view the actual course. Not sure if I'm missing something, but I would think we would want those numbers to be the same.

**5 Assessments**
screen shot 2016-01-05 at 11 51 30 am

**4 Assessments**
screen shot 2016-01-05 at 11 51 42 am


2) I'm using selectize.js to make the select look less awful. Selectize takes an option `allowEmptyOption`, which should ignore your `include_blank` option, but it's removing it. The [documentation](https://github.com/selectize/selectize.js/blob/master/docs/usage.md) says:

>`allowEmptyOption` - If true, Selectize will treat any options with a "" value like normal.

I'm not sure if it's removing it because `value` is nothing rather than `value=""`. Any ideas?

derekprior:
I'll take a look at both of those and merge this in when I'm done.

app/assets/stylesheets/base/_tables.scss
```diff
@@ -41,21 +42,16 @@ th {
...
+  width: 300px;
+}
+
 caption {
   margin: $small-spacing 0;
   text-align: left;
   width: 75%;
 }
 
-.table-section-header {
```
alexandra-aas: @derekprior This is actually being used for the activity feed now, just FYI for when you merge because it's being removed here and then used in another branch.

derekprior: Thanks. I'll add it back.

derekprior:
@alexandra-aas I added a trello card for the count issue. It's not super straight forward, but it is a bug, give the way we are now displaying assessments. It's just not something I think should hold up this PR right now. See: https://trello.com/c/ik4C01JE

The fix for the second bug was to roll back to electize 0.11.2. 0.12.x introduced a bug in the `allowEmptyOption` behavior.

PR #141, opened 2016-01-06
Redirect if only one course or subject

alexandra-aas:


alexandra-aas:
@derekprior Can you take a look at the above? Also, the tests I changed assume the user has one department, should I add coverage for the cases where the user has multiple departments?

app/controllers/manage_assessments/dashboard_controller.rb
```diff
@@ -2,5 +2,9 @@ class ManageAssessments::DashboardController < ApplicationController
   def show
     @courses = policy_scope(Course.includes(:outcomes_with_metadata))
     authorize(:generic, :view_assessments?)
+
+    if @courses.count == 1
```
derekprior: We should probably use `.length` here instead of `.count`. The difference is subtle and small in this case, but `.count` will issue a query returning only the count of records, while `.length` will get all of the records and then check the length of the array.

We need all of the records whether the count is 0, 1, or 10, so we might as well just load it up to avoid doing the `count` query and then the full query immediately after.

app/controllers/manage_results/subjects_controller.rb
```diff
@@ -3,6 +3,10 @@ def index
     @subjects = policy_scope(Subject).
       with_direct_assessments.
       sorted_by_number
+
+    if @subjects.count == 1
```
derekprior: Ditto above here.

app/controllers/manage_results/subjects_controller.rb
```diff
@@ -3,6 +3,10 @@ def index
...
+      redirect_to action: "show", id: @subjects.first.id
```
derekprior: We should use the path helper here instead: `manage_results_subject_path(@subjects.first.id)`

derekprior:
This looks good. Just a few minor points.

> should I add coverage for the cases where the user has multiple departments?

Hmm. I'm not sure. Technically, we could test the redirects with controller specs instead of feature specs and then leave the feature specs testing the interface where there are multiple courses/subjects. Let me take a quick look at how hard this would be to setup.

derekprior:
Actually, I'm okay with these specific tests skipping that step. Don't worry about the coverage.

alexandra-aas:
@derekprior Good to merge then?

derekprior:
:+1:

PR #142, opened 2016-01-06
Make logo crisp

alexandra-aas:
Not really necessary, but nice for people on high res laptops. Before and After (have to click the image to expand it and see the difference):

screen shot 2016-01-06 at 3 10 11 pm


derekprior:
:+1:

PR #144, opened 2016-01-06
Fix period dropdowns for new result

alexandra-aas:
**Before**
screen shot 2016-01-06 at 4 34 56 pm

**After**
screen shot 2016-01-06 at 4 34 38 pm


derekprior:
:+1: - I meant to fix this when I saw it yesterday. Thanks!

PR #145, opened 2016-01-06
View assessments for outcomes

alexandra-aas:
Easily view assessments for an outcome for a course:

view_assessments_for_course

### DIFF
--- a/app/views/manage_outcomes/outcomes/_outcomes.html.erb
+++ b/app/views/manage_outcomes/outcomes/_outcomes.html.erb
@@ -13,7 +13,13 @@
         <td><%= render "progress_bar",
           amount: outcome.active_assessments_with_results_count,
           total: outcome.active_assessments_count %></td>
-        <td><%= link_to t(".edit_outcome"), edit_manage_outcomes_outcome_path(outcome) %></td>
+        <td>
+          <%= link_to t(".edit_outcome"), edit_manage_outcomes_outcome_path(outcome) %>
+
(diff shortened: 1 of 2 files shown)


app/views/manage_outcomes/outcomes/_outcomes.html.erb
```diff
@@ -13,7 +13,13 @@
         <td><%= render "progress_bar",
           amount: outcome.active_assessments_with_results_count,
           total: outcome.active_assessments_count %></td>
-        <td><%= link_to t(".edit_outcome"), edit_manage_outcomes_outcome_path(outcome) %></td>
+        <td>
+          <%= link_to t(".edit_outcome"), edit_manage_outcomes_outcome_path(outcome) %>
+
+          <% if outcome.outcome_assessments.any? %>
+            <%= link_to t(".view_assessments"), manage_assessments_course_assessments_path(@course, outcome_ids: outcome) %>
```
alexandra-aas: @derekprior Using this to filter when coming from a link. I noticed when you filter direcrtly on the page you get `?utf8=✓&outcome_ids=2`. Is the utf param something I should include here as well?

derekprior: No, it's not really necessary in this context. I should update the form not to send it when filtering on the assessments screen as well.

derekprior:
This is great. Everything tying together. Thanks!

PR #147, opened 2016-01-07
Activity stats

alexandra-aas:
- Derek needs to add actual numbers for stats

screencapture-localhost-3000-activities-1452187856351

### DIFF
--- a/app/assets/stylesheets/_activity_stat.scss
+++ b/app/assets/stylesheets/_activity_stat.scss
@@ -1,0 +1,50 @@
+.activity-stats {
+  display: flex;
+  flex-direction: column;
+  margin-bottom: $large-spacing;
+
+  @include media($medium-screen-up) {
+    flex-direction: row;
+  }
+}
+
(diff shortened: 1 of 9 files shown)


alexandra-aas:
@derekprior Not sure if you'll get around to this, but ready to go if you do.

derekprior:
@jdaries is there a way we can know the current semester give a date?

jdaries:
@derekprior we can make a guess, but may be slightly off due to shifts in day of week, calendar. I think in general, we could do the following:
Jan: JA
Feb through and including May: SP
June July August: SU
September through December: FA

alexandra-aas:
@jdaries If there are other stats that seem more relevant we can use those instead. Just thought these might be helpful to Stephanie to have a more overarching idea of how on target they are for people adding data.

jdaries:
@alexandra-aas i think those are good, but i realized that since SU isn't a choice for the results (we don't really have summer classes) then we should extend the SP dates through August

PR #148, opened 2016-01-07
Add search functionality

alexandra-aas:
- Derek needs to wire up


alexandra-aas:
@derekprior This is ready for you if you get to it. I'm assuming you'll probably want to split this into two tickets since it is for two distinct views, but I wanted to make sure I got all the styling in there for you, so it works as expected in both places.

PR #152, opened 2016-01-07
Update documentation

alexandra-aas:


alexandra-aas:
@jdaries Mind reading through these quick to double check they align with the new flows?

PR #154, opened 2016-01-12
Conditionally display "View Histogram" link

derekprior:
We only want to show the gradebook histogram link if we have access to the
gradebook. If we can't get the gradebook for any reason, then don't show
the link.


derekprior:
Added some code to remove the possibly previously existing histogram link when the user changes the period so no stale link is displayed when we're checking to see if the gradebook is available.

PR #156, opened 2017-05-02
Add API to get outcomes for a subject

derekprior:
Faculty want to be able to display outcomes associated with a
class/subject on their own websites. This API will allow them to do
that.

The API format is extremely simplistic at this point. Jon is checking to
see if there are preferred standards (JSON API?) within use at MIT. In
the meantime, I thought this simplified format would likely be of the
most use to the people we think will be consuming this API.

Example output:

```
[
    {
        "description": "an ability to function on multidisciplinary teams",
        "name": "d"
    },
    {
        "description": "an ability to identify, formulate, and solve engineering problems",
        "name": "e"
    }
]
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 ruby "2.2.2"
 
 gem "rails", "~> 5.0.0"
+gem "active_model_serializers", "~> 0.10.6"
 gem "email_validator"
 gem "flutie"
 gem "high_voltage"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,6 +3,12 @@ Rails.application.routes.draw do
     resources :results, only: [:new, :create]
   end
 
+  namespace :api do
+    resources :subjects, only: [], id: /[A-Z0-9\.]+?/i do
+      resources :outcomes, only: [:index]
+    end
+  end
+
   namespace :manage_assessments do
     root "dashboard#show"
 
--- a/spec/requests/api/subject_outcomes_spec.rb
+++ b/spec/requests/api/subject_outcomes_spec.rb
@@ -1,0 +1,22 @@
+require "rails_helper"
+
+describe "subject outcomes api" do
+  it "returns a unique list of outcomes for the given subject" do
+    subject = create(:subject, number: "1.234")
+    outcome,_ = create_pair(:outcome)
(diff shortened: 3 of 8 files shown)


config/routes.rb
```diff
@@ -3,6 +3,12 @@
     resources :results, only: [:new, :create]
   end
 
+  namespace :api do
+    resources :subjects, only: [], id: /[A-Z0-9\.]+?/i do
```
derekprior: The agreed upon identifier for a subject is its number, which looks something like 6.100A or 2.URG. Without overriding the id format here, Rails was not allowing the dot.

Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 ruby "2.2.2"
 
 gem "rails", "~> 5.0.0"
+gem "active_model_serializers"
```
geoffharcourt: Since there are big API changes between versions, I think this should be pessimistically locked.

spec/requests/api/subject_outcomes_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+describe "subject outcomes api" do
+  it "returns a unique list of outcomes for the given subject" do
+    subject = create(:subject, number: '1.234')
+    outcome,_ = create_pair(:outcome)
+    assessment = create_pair :direct_assessment,
+      subject: subject,
+      outcomes: [outcome]
```
geoffharcourt: Parens for assignment?

spec/requests/api/subject_outcomes_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+describe "subject outcomes api" do
+  it "returns a unique list of outcomes for the given subject" do
+    subject = create(:subject, number: '1.234')
```
geoffharcourt: Nitpick: elsewhere it looks like double quotes are used.

spec/requests/api/subject_outcomes_spec.rb
derekprior: I like the style here for factories (name on line with method call, hash elements on their own lines). Adding parens makes it not as nice looking and confuses vim. The thoughtbot systle guide (each argument on its own line) would keep vim happy but this is a place where I like to stray.

PR #160, opened 2017-05-09
Allow cross origin requests for class outcomes

derekprior:
API users would like to request a list of outcomes associated with the
given class (AKA subject) from an AJAX request on their class website.
To enable this, the API needs to accept requests from domains other than
the current host. We use Rack::Cors to allow get requests to the API
from any domain.

CORS: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS


config/initializers/cors.rb
```diff
@@ -0,0 +1,6 @@
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  allow do
+    origins '*'
```
thorncp: Other strings in this PR use double quotes, should this file also use them?

config/initializers/cors.rb
```diff
@@ -0,0 +1,6 @@
...
+    resource '/api/*', headers: :any, methods: %i(get)
```
mike-burns: `%i(get)` is one character *more* than `[:get]`!

config/initializers/cors.rb
derekprior: Yup. Thanks. Last client used single quotes so I'm weaning myself off of it.

config/initializers/cors.rb
derekprior: I'm okay with that. It's really going to pay character saving dividends if we add `post`, `patch`, `put` and `delete`. Just think of all those saved `:` and `,`!

PR #162, opened 2017-05-10
Populate subjects from the subject listings API

derekprior:
We were previously populating subjects from an in-repo CSV, which we
imported with a rake task. MIT offers the [subject listings API][1] as
the official source for this information.

This PR introduces a client for this API and converts the existing rake
task import to use the client as the source of data. The client defaults
to using a "fake" adapter which loads from the json fixture in the app.
If the app is configured with access keys for the API, the real API will
be used.

[1]:https://developer.mit.edu/apis-subjectlistings


db/seeds.rb
```diff
@@ -53,6 +53,5 @@
     )
   end
 
-  subject_csv = Rails.root.join("config", "subjects.csv")
-  SubjectImporter.new(subject_csv).run
+  Rake::Task["subjects:import"].invoke("2017FA")
```
margaretoti: This hardcodes in "2017FA". Will the semester need to be changed regularly?

margaretoti: This hardcodes in "2017FA". Will the semester need to be changed regularly?

derekprior: I'm not crazy about this either, but I think it's relatively harmless. Seeding with old data wouldn't be bad -- we can run the rake task again with the correct term.

That said, I wonder if we can be smart about it somehow. Obviously, we can get the current year, maybe we can also divide the months up into SP/JA/SU/FA semesters?

margaretoti: Dividing up the months could work. Although, you are right that it is easy to run the rake task again with the correct term. Let's leave things as is for now. We can make a new PR for this and tackle during the end if there is time.

PR #165, opened 2017-05-15
Add nickname field to outcomes

derekprior:
Associating a nickname to each outcome will allow us to concisely
represent outcomes in the UI in a way that the user can still understand
what the outcomes mean. The nickname must be unique to the associated
course.

We updated all existing outcomes with a nickname. Outcomes based on
standard outcomes were updated to use the canonical nicknames provided
by Jon. Custom outcomes were updated to use their name as nickname. Jon
will investigate updating these manually at a later time.

Finally, we opened a new Trello card to cover displaying the nickname
alongside the outcomes in the "Manage Outcomes" views as we weren't sure
how to best do this right now.


dgalarza:
LGTM 🚢

PR #166, opened 2017-05-16
Add outcome coverage for a class to a course

margaretoti:
[Trello Card](https://trello.com/c/UxUB8m5r)

This branch adds outcome coverage to a subject/class. Outcome coverage refers to the one or more outcomes covered by any given subject/class in a course. Connecting outcomes to subjects/classes and then subjects/classes to courses means that the UI can show a user which outcomes have been matched to a class in the course and which outcomes still need to be demonstrated by a subject/class in the course.


app/models/outcome_coverage.rb
```diff
@@ -0,0 +1,8 @@
+class OutcomeCoverage < ActiveRecord::Base
+  belongs_to :course
+  belongs_to :outcome
+  belongs_to :subject
+
+  delegate :title, to: :subject, prefix: true
+  delegate :name, :nickname, to: :outcome, prefix: true
```
geoffharcourt: worth :abc:-ing these by the thing they delegate to?

config/routes.rb
```diff
@@ -9,6 +9,7 @@
     root "dashboard#show"
 
     resources :courses, only: [:index] do
+      resources :outcome_coverages, only: [:new, :create, :index]
```
geoffharcourt: :abc: these? (I have some trauma from my last app having a very long routes list).

spec/features/user_adds_outcome_coverage_to_a_course_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    subject = create(:subject)
+    course = create(:course)
+    outcome = create(:outcome, course: course)
+    user = user_with_assessments_access_to(course.department)
+
+    visit manage_assessments_course_outcome_coverages_path(course, as: user)
+    click_on I18n.t('manage_assessments.outcome_coverages.index.add_a_class')
+    select subject.title, from: "outcome_coverage[subject_id]"
+    select outcome.nickname, from: "outcome_coverage[outcome_id]"
+    click_button I18n.t('helpers.submit.outcome_coverage.create')
```
geoffharcourt: I18n is included in tests, so I think you can just do `t(...)` here.

https://github.com/MIT-IR/abet/blob/master/spec/support/I18n.rb

app/views/manage_assessments/outcome_coverages/new.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= simple_form_for @outcome_coverage, url: manage_assessments_course_outcome_coverages_path(course_id: @outcome_coverage.course.id) do |form| %>
```
derekprior: I think we can just do: `manage_assessments_course_outcome_coverages_path(@outcome_coverage.course)`

derekprior: woah? really. Sounds interesting. Disregard for now then.

margaretoti: Sure. Will do.

PR #167, opened 2017-05-16
Clean up type system partials to ease future design

mborsare:
The type system was using a couple of different partials. This PR
consolidates those two into one. This lessens duplication and creates
an easier to maintain system.


derekprior:
⛩

PR #168, opened 2017-05-16
Refactor outcome coverages some

derekprior:
* Moved `outcome_coverages#index` to `course#show` as it seems more
accurate that we're representing the state of the assessments for the
course than an index of coverages.
* Cleaned up the select boxes on the coverage form to have proper labels
in the correct order.
* Changed courses#show html to be more in line with what the final
designed product might use.

### DIFF
--- a/app/models/outcome.rb
+++ b/app/models/outcome.rb
@@ -23,4 +23,8 @@ class Outcome < ActiveRecord::Base
   def to_s
     "#{name} - #{description}"
   end
+
+  def label
+    "#{name} - #{nickname}"
+  end
 end
--- a/app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
+++ b/app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
@@ -1,4 +1,4 @@
-<h2><%= outcome_coverage.subject_title %></h2>
+<h2><%= outcome_coverage.subject %></h2>
 
 <div>
   <p>Outcome <%= outcome_coverage.outcome_name %></p>
--- a/config/locales/manage_assessments.en.yml
+++ b/config/locales/manage_assessments.en.yml
@@ -60,6 +60,6 @@ en:
         description: Description
         heading: Outcomes for %{name}
         label: Label
-    outcome_coverages:
-      index:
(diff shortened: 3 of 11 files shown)


config/locales/manage_assessments.en.yml
```diff
@@ -60,6 +60,6 @@ en:
         description: Description
         heading: Outcomes for %{name}
         label: Label
-    outcome_coverages:
-      index:
+      show:
         add_a_class: Add a class
+        course_number: Course %{number}
```
margaretoti: I like this. It explains why you were able to do this `<div><%= t(".course_number", number: @course.number) %></div>` in the view.

app/models/outcome.rb
```diff
@@ -23,4 +23,8 @@ class Outcome < ActiveRecord::Base
   def to_s
     "#{name} - #{description}"
   end
+
+  def label
+    "#{name} - #{nickname}"
```
margaretoti: Going based on Mike's suggestion, we can make the name capital so that it matches how it will display in the manage outcomes views.

```
def label
    "#{name.upcase} - #{nickname}"
end 
```

app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
```diff
@@ -1,4 +1,4 @@
-<h2><%= outcome_coverage.subject_title %></h2>
```
margaretoti: Should we now remove the corresponding delegate (`delegate :title, to: :subject, prefix: true`) in the outcome_coverage model as it is no longer being used?

app/models/outcome.rb
derekprior: This is a good idea.

If we want "Name" to be upper case everywhere, we're probably better off handling that in a way that makes sure it's upper case going into the database. There's a few different options we can use for that. I can follow up with another PR for it.

app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
derekprior: Yup. Thanks.

PR #169, opened 2017-05-16
Outcomes display with nicknames

margaretoti:
[Trello card](https://trello.com/c/xCtdJH7Z)
In manage outcomes views, all outcomes will be displayed with corresponding nicknames.

### DIFF
--- a/app/views/manage_outcomes/outcomes/_outcomes.html.erb
+++ b/app/views/manage_outcomes/outcomes/_outcomes.html.erb
@@ -9,7 +9,10 @@
   <tbody>
     <% course.outcomes_with_metadata.each do |outcome| %>
       <tr>
-        <td><%= outcome %></td>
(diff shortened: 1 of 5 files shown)


app/models/outcome_with_metadata.rb
```diff
@@ -1,3 +1,7 @@
 class OutcomeWithMetadata < Outcome
   self.table_name = "outcomes_with_metadata"
+
+  def nickname
```
derekprior: Is this needed? `OutcomeWithMetadata` is a view that already selects `outcomes.nickname`. So theoretically, `nickname` should already be a method on this class.

margaretoti: Good point. I didn't see `nickname` as a column and forgot that it inherits from `outcome` already.

PR #170, opened 2017-05-16
Rename Outcome name to label and force uppercase

derekprior:
This is less ambiguous than `name`. This change also upcases all stored
labels and sets the max length of a label to 5 characters.

The type coerces any label values to uppercase both in queries and in
inserts. For example, `StandardOutcome.where(label: 'a')` will search
for standard outcomes with a label of `A`. `Outcome.create(label: 'z')` will
similarly create an outcome with a label of `Z`.

This allows us to normalize the case at the app level.

### DIFF
--- a/app/models/outcome.rb
+++ b/app/models/outcome.rb
@@ -1,4 +1,6 @@
 class Outcome < ActiveRecord::Base
+  attribute :label, Label.new
+
   belongs_to :course, counter_cache: true
 
   has_one :department, through: :course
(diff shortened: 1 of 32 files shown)


app/models/outcome.rb
```diff
@@ -14,17 +16,18 @@ class Outcome < ActiveRecord::Base
     reject_if: ->(attributes) { attributes[:level].blank? },
     allow_destroy: true
 
-  validates :name, presence: true, uniqueness: { scope: :course_id }
+  validates :label, presence: true, uniqueness: { scope: :course_id },
+    length: { maximum: 500 }
```
margaretoti: Should the maximum be 5 characters? The schema reads `t.string   "label", limit: 5, null: false`.

derekprior: Yup. Good catch. Thanks!

PR #171, opened 2017-05-16
High voltage design for when a course has no assessments

mborsare:
**small screen**
screen shot 2017-05-16 at 5 08 28 pm

**big screen**
hover

### DIFF
--- a/config/locales/manage_assessments.en.yml
+++ b/config/locales/manage_assessments.en.yml
@@ -65,6 +65,9 @@ en:
         description: Description
         heading: Outcomes for %{name}
         label: Label
-      show:
+      show: &courses_show
         add_a_class: Add a class
         course_number: Course %{number}
+      show_without_coverages:
+        <<: *courses_show
+        no_classes: "%{name} doesn’t have any classes yet"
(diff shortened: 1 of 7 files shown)


app/views/pages/course-no-classes.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <h2 class="headline-poster headline-poster-copy narrow-content">
+    Aerospace Engineering doesn’t have any classes yet
+  </h2>
+</div>
+
+<%= link_to "Add a class", "#!", class: "headline-poster headline-poster-action" %>
+
+<script type="text/javascript">
+  $(function() {
+    $(".headline-poster-action").mouseenter(function(){
```
ryancoughlin: Space between parens `() {`

app/views/pages/course-no-classes.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      $("body").addClass('headline-poster-takeover');
+    });
+    $(".headline-poster-action").mouseleave(function(){
```
ryancoughlin: You won't notice this much in this case above, but it's best to cache selectors (this is so lightweight though):

`var $headlinePosterAction = $(".headline-poster-action");`

config/locales/manage_assessments.en.yml
```diff
@@ -65,6 +65,9 @@ en:
         description: Description
         heading: Outcomes for %{name}
         label: Label
-      show:
+      show: &courses_show
```
margaretoti: Out of curiosity, why does `show:` need to be aliased to `courses_show`?

mborsare:
LGTM!

PR #172, opened 2017-05-17
Validate outcome and subject on outcome coverage

derekprior:
Users could previously create a coverage without any of the required
association. This updates the data model to require the associations and
then updates the model to add appropriate user-facing validations.

Validations were not added to the `course` association because it is not
user-facing. If we somehow attempt to create a coverage without a
course, this is a programming error and should result in an application
error.

### DIFF
--- a/app/models/outcome_coverage.rb
+++ b/app/models/outcome_coverage.rb
@@ -1,7 +1,10 @@
 class OutcomeCoverage < ActiveRecord::Base
   belongs_to :course
-  belongs_to :outcome
-  belongs_to :subject
+  belongs_to :outcome, required: true
+  belongs_to :subject, required: true
+
+  validates :outcome_id, presence: true
+  validates :subject_id, presence: true
 
   delegate :label, :nickname, to: :outcome, prefix: true
 end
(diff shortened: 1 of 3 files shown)


app/models/outcome_coverage.rb
```diff
@@ -1,7 +1,10 @@
 class OutcomeCoverage < ActiveRecord::Base
   belongs_to :course
-  belongs_to :outcome
-  belongs_to :subject
+  belongs_to :outcome, required: true
+  belongs_to :subject, required: true
```
margaretoti: This is simply a comment/ fun fact that I never knew: Apparently in Rails 5, `required: true` is the default but in Rails 4.x `required: false` was the default, which is why you needed the line `validates, presence: true`. See [here](https://jasoncharnes.com/2016/03/29/rails-belongs_to-required-in-rails-5/) or [here](http://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html).

derekprior: Yeah, it's the default for Rails 5 generated apps, but when you update from Rails 4 it adds this: https://github.com/MIT-IR/abet/blob/master/config/initializers/new_framework_defaults.rb

That maintains the Rails 4 behavior. I tried removing that when we upgraded to Rails 5 but I got a bunch of test failures I didn't want to look into. That could be something we do when we're looking for an activity to round out a day.

PR #174, opened 2017-05-17
Rename OutcomeCoverage to Coverage

derekprior:
`Coverage` adequately describes what we're capturing and also frees up
the name `OutcomeCoverage` for use when we support coverage of many
outcomes in a single subject.


derekprior:
>It's perfect.

Time to retire!

high note

PR #176, opened 2017-05-17
Enable multiple outcomes per coverage

derekprior:
This allows a given class to say that it covers multiple outcomes. We do
this by adding an `OutcomeCoverage` association to `Coverage`, removing
the previously existing `belongs_to :outcome` there.

In this change, I updated the controllers, models, and tests to allow
for the possibility of multiple outcomes, but the UI does not yet expose
the ability to add a second outcome. That will come in a future change.

### DIFF
--- a/app/models/coverage.rb
+++ b/app/models/coverage.rb
@@ -1,10 +1,28 @@
 class Coverage < ActiveRecord::Base
   belongs_to :course
-  belongs_to :outcome, required: true
   belongs_to :subject, required: true
 
-  validates :outcome_id, presence: true
-  validates :subject_id, presence: true
+  has_many :outcome_coverages
+  has_many :outcomes, through: :outcome_coverages
 
-  delegate :label, :nickname, to: :outcome, prefix: true
+  accepts_nested_attributes_for :outcome_coverages,
+    allow_destroy: true,
+    reject_if: :all_blank
+
+  validates :subject_id, presence: true
+  validate :must_have_outcomes
+
+  private
+
+  def must_have_outcomes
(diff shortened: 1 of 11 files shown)


app/models/coverage.rb
```diff
@@ -1,10 +1,28 @@
...
-  validates :outcome_id, presence: true
+  has_many :outcome_coverages
+  has_many :outcomes, through: :outcome_coverages
+
+  accepts_nested_attributes_for :outcome_coverages,
+    allow_destroy: true,
+    reject_if: :all_blank
+
   validates :subject_id, presence: true
+  validate :must_have_outcomes
```
margaretoti: Nice!

app/views/manage_assessments/coverages/new.html.erb
```diff
@@ -1,5 +1,14 @@
...
-  <%= form.input :outcome_id, collection: @coverage.course.outcomes, label_method: :to_short_s %>
+  <%= form.input :subject_id,
+    collection: Subject.sorted_by_number,
+    label_method: :to_s %>
+
+  <%= form.simple_fields_for :outcome_coverages do |fields| %>
+    <%= fields.input :outcome_id,
+      collection: @coverage.course.outcomes,
+      label_method: :to_short_s,
+      required: fields.options[:child_index] == 0 %>
```
margaretoti: Do you mind explaining what this line is doing?
`required: fields.options[:child_index] == 0`

derekprior: when you call `fields_for`, the form build object keeps track of which item in the array of associated objects you are on. This sets the `outcome_id` field to be `required: true` when we're rendering the 0th index (or the first item).

PR #178, opened 2017-05-17
Add JavaScript testing of coverage creation

derekprior:
We will be adding some JavaScript to this page in order to support
adding many outcomes to a coverage dynamically. To get adequate test
coverage on this, we need to have a JavaScript test driver.

We usually use capybara-webkit for this, but QT was giving me troubles
and the world seems to be moving on to chromedriver. I decided to give
chromedriver a try here. Installation was an absolute picnic compared to
QT and capybara-webkit.

With the test case now running in a JavaScript driver, we can no longer
faithfully interact with the select boxes, which are actually hidden by
Selectize. I updated the test to interact with the Selectize input
instead.

### DIFF
--- a/spec/features/user_adds_coverage_to_a_course_spec.rb
+++ b/spec/features/user_adds_coverage_to_a_course_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 feature "user adds coverage to a course" do
-  scenario "successfully" do
+  scenario "successfully", js: true do
     subject = create(:subject)
     course = create(:course)
     outcome = create(:outcome, course: course)
@@ -9,13 +9,19 @@ feature "user adds coverage to a course" do
 
     visit manage_assessments_course_path(course, as: user)
     click_on t('manage_assessments.courses.show.add_a_class')
-    select subject.title, from: "Subject"
-    select outcome.nickname, from: "Outcome"
+    selectize subject.title, from: "Subject"
+    selectize outcome.nickname, from: "Outcome"
     click_button t('helpers.submit.coverage.create')
 
     within("#matched_outcomes") do
       expect(page).to have_content(subject.title)
       expect(page).to have_content(outcome.nickname)
     end
+  end
+
+  def selectize(item, from:)
+    container = find_field(from, visible: false).first(:xpath, ".//..")
+    container.find(".selectize-control").click
+    container.find("div.option", text: item).click
   end
 end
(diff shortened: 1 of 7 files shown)


spec/features/user_adds_coverage_to_a_course_spec.rb
```diff
@@ -1,21 +1,27 @@
...
+    selectize outcome.nickname, from: "Outcome"
     click_button t('helpers.submit.coverage.create')
 
     within("#matched_outcomes") do
       expect(page).to have_content(subject.title)
       expect(page).to have_content(outcome.nickname)
     end
   end
+
+  def selectize(item, from:)
```
derekprior: Is all of this too much? This is how I replicate selecting from a selectize'd input. Should I use selectize's javascript api instead (via execute_script)?

spec/features/user_adds_coverage_to_a_course_spec.rb
```diff
@@ -1,21 +1,27 @@
...
+    container = find_field(from, visible: false).first(:xpath, ".//..")
```
calebthompson: visible: false scares me. Do we know that a user can click on it and it's not behind some div capturing clicks?

calebthompson:
lgtm

spec/features/user_adds_coverage_to_a_course_spec.rb
margaretoti: Perhaps using the selectize javascript api is a good idea. I foresee needing to use it to implement the nice form designs in the mockups. That said, I am not 100 % sure as I do not think that using selectize will necessarily make the code cleaner. From the documentation, it looks like you need to call quite a few functions. Maybe we can look at this together; I am curious to see what you had in mind if the selectize api is used.

spec/features/user_adds_coverage_to_a_course_spec.rb
derekprior: The user can't click on it. This is the real `select` box, which `selectize` hides with it's fancy select box. In order to interact with the correct `selectize` box, I find the hidden select and then interact with visible elements that are in the same container.

spec/features/user_adds_coverage_to_a_course_spec.rb
derekprior: I looked into this a bit, and I think the only things that would change would be the 2nd and 3rd lines of this method -- we'd be able to replace those with `page.driver.execute_script` executing some javascript to set the value of the selectize. It doesn't look like it'd be much simpler than what we have here and what we have here is a bit more true to how a user would interact with it. So I think we should go with this for now. Can revisit if this gives us problems.

PR #179, opened 2017-05-17
Design for the class card component

mborsare:
This is a scaffold. There are internal components that will be built down
the line.

**mobile**
screen shot 2017-05-17 at 4 27 07 pm

**desk**
screen shot 2017-05-17 at 4 27 12 pm

https://trello.com/c/tiXncRa3

### DIFF
--- a/app/assets/stylesheets/base/_variables.scss
+++ b/app/assets/stylesheets/base/_variables.scss
@@ -41,8 +41,9 @@ $light-blue: aliceblue;
 $light-blue-border: shade($light-blue, 5%);
 $dark-gray: #484343;
 $brown: #645858;
-$medium-gray: #aaa;
+$medium-gray: tint($brown, 20%);
 $light-gray: #ddd;
+$lighter-gray: #f7f7f7;
 $lightest-gray: #fafafa;
 $white: #fff;
 $required-color: #f58122;
@@ -62,7 +63,10 @@ $base-letter-spacing: 1px;
 
 // Border
 $base-border-color: $light-gray;
+$lighter-border-color: $lighter-gray;
 $base-border: 1px solid $base-border-color;
+$thick-border-width: 2px;
+$thickest-border-width: 6px;
 
 // Forms
 $form-box-shadow: inset 0 1px 3px rgba(#000, 0.06);
@@ -85,6 +89,7 @@ $breadcrumb-color-active: $breadcrumb-color;
 
 // Transition
 $base-transition-duration: 0.2s;
+$longest-transition-duration: 1s;
 $base-transition: all $base-transition-duration ease-in-out;
 $long-transition: all ($base-transition-duration * 2) ease-in-out;
 
--- a/app/assets/stylesheets/components/_class-card.scss
+++ b/app/assets/stylesheets/components/_class-card.scss
@@ -1,0 +1,35 @@
+.class-card {
+  @include margin($base-spacing null);
+  background-color: $white;
+  border: $thickest-border-width solid $lighter-border-color;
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/base/_variables.scss
```diff
@@ -41,8 +41,9 @@ $light-blue: aliceblue;
 $light-blue-border: shade($light-blue, 5%);
 $dark-gray: #484343;
 $brown: #645858;
-$medium-gray: #aaa;
+$medium-gray: tint($brown, 20%);
```
alexandra-aas: Should this be `$medium-brown` then? Such a nice color name.

mborsare:
@margaretoti we broke test on this one. Would you mind taking a look?

app/assets/stylesheets/base/_variables.scss
```diff
@@ -41,8 +41,9 @@ $light-blue: aliceblue;
...
 $light-gray: #ddd;
+$black-haze: #f7f7f7;
```
alexandra-aas: Is this color name to avoid renaming the grays? I assumed it would be almost black until I saw the hex value.

app/assets/stylesheets/components/_class-card.scss
```diff
@@ -0,0 +1,37 @@
+$class-card-padding-y: $small-spacing;
+$class-card-padding-2y: $base-spacing;
+
+.class-card {
+  @include margin($base-spacing null);
+  border: $thickest-border-width solid $black-haze;
```
alexandra-aas: Do you want to map `$black-haze` to a border variable?

app/assets/stylesheets/components/_class-card.scss
```diff
@@ -0,0 +1,37 @@
...
+  padding: $class-card-padding-2y;
```
alexandra-aas: Not sure what the need is for the `y`? At first I thought it was because it could only be applied vertically, but in this case it's also used on the horizontal.

app/models/subject.rb
```diff
@@ -11,8 +11,4 @@ def self.sorted_by_number
   def self.with_assessments
     joins(:assessments).distinct
   end
-
-  def to_s
```
derekprior: We're still relying on this elsewhere, so I think we still need this method. This is causing the test failures.

mborsare: Yes, indeed. I checked it out from master and all is well.

alexandra-aas: @derekprior This shouldn't be fixed in here, but isn't `to_s` a built in rails method? It seems odd to me to name a custom method the same as one that's build in.

app/assets/stylesheets/base/_variables.scss
mborsare: I think it's okay to leave this one. The color reads as a warm gray.

app/assets/stylesheets/base/_variables.scss
mborsare: This is straight from Sip. I'll revisit it.

app/assets/stylesheets/components/_class-card.scss
mborsare: Ah yes. That was the intention, but I overlooked the declaration on line 8. Perhaps it's too much for such a small component?

app/models/subject.rb
derekprior: It is inherited from `Object`, yes. You are meant to override it with a valid string representation of the object. That's often useful in debugging and ocasionally other situations, but we're likely overusing `to_s` in the views.

PR #181, opened 2017-05-18
Allow multiple outcomes per coverage in the UI

derekprior:
Users can now add multiple outcomes to a new coverage in the UI. This
relies on previous work done to support this in the data model. This
isn't final UI or UX, but the functionality is at least now present.

multi-outcome

### DIFF
--- a/app/views/manage_assessments/coverages/new.html.erb
+++ b/app/views/manage_assessments/coverages/new.html.erb
@@ -23,5 +23,12 @@
       required: fields.options[:child_index] == 0 %>
   <% end %>
 
+  <div class="nested-form-actions">
+    <%= link_to_add_association t('.add_outcome'),
+      form,
+      :outcome_coverages,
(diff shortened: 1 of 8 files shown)


app/views/manage_assessments/coverages/new.html.erb
```diff
@@ -23,5 +23,12 @@
       required: fields.options[:child_index] == 0 %>
   <% end %>
 
+  <div class="nested-form-actions">
+    <%= link_to_add_association t('.add_outcome'),
+      form,
+      :outcome_coverages,
```
margaretoti: Cocoon is magic. 😄  It took me a while to figure out how it knew to render `app/views/manage_assessments/coverages/_outcome_coverage_fields.html.erb`.

derekprior: Yeah, it's really pretty nice.

PR #182, opened 2017-05-18
Add assignment to outcome coverage

margaretoti:
[Trello card](https://trello.com/c/zTh1BDd2)
A user can now add an assignment to specific outcome coverages. Each outcome coverage can only have one assignment associated with it. Currently, assignments can only be added; they cannot be edited or deleted. 

screen shot 2017-05-23 at 11 45 31 am

### DIFF
--- a/app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
+++ b/app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
@@ -4,3 +4,10 @@
 <p class="class-card-outcome class-card-outcome-nickname">
   <%= outcome_coverage.outcome_nickname %>
 </p>
+<% if outcome_coverage.assignment.present? %>
+  <p> In <%= outcome_coverage.assignment.name %> on <%= outcome_coverage.assignment.problem %>
+<% else %>
+  <p class="class-card-outcome">
+    <%= link_to t(".add_assignment"), new_manage_assessments_outcome_coverage_assignment_path(outcome_coverage) %>
+  </p>
+<% end %>
--- a/app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
+++ b/app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
@@ -5,7 +5,11 @@
   <%= outcome_coverage.outcome_nickname %>
 </p>
 <% if outcome_coverage.assignment.present? %>
-  <p> In <%= outcome_coverage.assignment.name %> on <%= outcome_coverage.assignment.problem %>
+  <p>
+    <%= t(".assignment_name", name: outcome_coverage.assignment.name) %>
+    <%= t(".assignment_problem",
+        problem: outcome_coverage.assignment.problem) if outcome_coverage.assignment.problem.present? %>
+  </p>
 <% else %>
   <p class="class-card-outcome">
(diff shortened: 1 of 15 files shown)


app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
```diff
@@ -4,3 +4,10 @@
 <p class="class-card-outcome class-card-outcome-nickname">
   <%= outcome_coverage.outcome_nickname %>
 </p>
+<% if outcome_coverage.assignment.present? %>
+  <p> In <%= outcome_coverage.assignment.name %> on <%= outcome_coverage.assignment.problem %>
```
derekprior: What do you think about using I18n for this?

`t(".assignment", name: outcome_coverage.name, problem: outcome_coverage.problem)`

db/migrate/20170518162321_create_assignments.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAssignments < ActiveRecord::Migration[5.0]
+  def change
+    create_table :assignments do |t|
+      t.timestamps null: false
+      t.references :outcome_coverage, index: true, foreign_key: true, null: false
+      t.string :name, null: false
+      t.string :problem, null: false
```
derekprior: From the mockup, I think this is supposed to be optional? https://marvelapp.com/8b0f34e/screen/27979251

db/migrate/20170518162321_create_assignments.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAssignments < ActiveRecord::Migration[5.0]
+  def change
+    create_table :assignments do |t|
+      t.timestamps null: false
+      t.references :outcome_coverage, index: true, foreign_key: true, null: false
+      t.string :name, null: false
```
derekprior: We should probably have a validation for this in the model as well.

app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
```diff
@@ -4,3 +4,14 @@
 <p class="class-card-outcome class-card-outcome-nickname">
   <%= outcome_coverage.outcome_nickname %>
 </p>
+<% if outcome_coverage.assignment.present? %>
+  <p>
+    <%= t(".assignment_name", name: outcome_coverage.assignment.name) %>
+    <%= t(".assignment_problem",
+        problem: outcome_coverage.assignment.problem) if outcome_coverage.assignment.problem.present? %>
```
margaretoti: Now that problem is optional we only want to display the text up to the word `on` when a problem is not given. 
screen shot 2017-05-23 at 1 48 56 pm

spec/models/assignment_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+describe Assignment do
+  describe "validations" do
+    it { should validate_presence_of(:name) }
+  end
+
+  describe "associations" do
+    it { should belong_to(:outcome_coverage) }
+    it { should have_one(:outcome).through(:outcome_coverage) }
+  end
+end
```
margaretoti: These tests might be not be needed. I noticed that `assessment_spec.rb` tested the validations, so I added these tests to be consistent. Happy to remove them.

derekprior: They're fine. I typically don't add them, but I also don't have a problem with other people adding them. They don't add much overhead.

PR #184, opened 2017-05-22
Rename Gradebook::Assignments

margaretoti:
This branch renames `Gradebook::Assignments` to `Gradebook::GradebookAssignments`. This change makes it possible to namespace the `Assignments` model and controller under
`Manage_Assessments`. This new `Assignments` will replace the existing `Assessments` model and controller currently nested under`Manage_Assessements`.


derekprior:
do we also need to do this with `lib/gradebook_client/assignment` ?

derekprior:
Or can that co-exist with `app/models/assignment` ?

config/routes.rb
```diff
@@ -43,9 +43,9 @@
...
       as: :gradebook
 
-    get "subjects/:subject_id/:semester/:year/assignments",
-      to: "assignments#index",
-      as: :assignments
+    get "subjects/:subject_id/:semester/:year/gradebook_assignments",
+      to: "gradebook_assignments#index",
+      as: :gradebook_assignments
 
     resources :assignments, only: [] do
```
margaretoti: @derekprior, I think you are correct. I went in and changed all the lib files. Things get a bit weird when I rename line 50 to `resources :gradebook_assignments, only: [] do`. But I will take another look.

PR #188, opened 2017-05-24
Design for the add assignment madlib form

mborsare:
This PR makes the add a class form match the styles established in the
add a class/outcome form.

add assignment

https://trello.com/c/QW12ljja

### DIFF
--- a/app/views/manage_assessments/assignments/new.html.erb
+++ b/app/views/manage_assessments/assignments/new.html.erb
@@ -1,5 +1,21 @@
-<%= simple_form_for @assignment, url: manage_assessments_outcome_coverage_assignments_path do |form| %>
-  <%= form.input :name, label: t(".assignment") %>
-  <%= form.input :problem, label: t(".problem") %>
-  <%= form.submit t(".add") %>
+<%= content_for :full_bleed do %>
+  <div class="madlib-header">
+    <h1 class="headline-narrow">
+      <%= t(".headline") %>
+    </h1>
+    <h2 class="headline-narrow-subhead">
+      <%= @assignment.course_number %>
+      &middot;
+      <%= @assignment.course_name %>
+    </h2>
+  </div>
+<% end %>
+
+
+<%= simple_form_for @assignment, html: { class: "madlib" }, url: manage_assessments_outcome_coverage_assignments_path do |form| %>
+  <%= form.input :name %>
+  <%= form.input :problem %>
+  <%= form.submit %>
+
+  <%= link_to t("helpers.cancel_button"), :back %>
 <% end %>
--- a/app/assets/javascripts/input_marking.js
+++ b/app/assets/javascripts/input_marking.js
@@ -1,0 +1,20 @@
+$(function(){
+  var markInput = function() {
+    input = $(this)
+    value = $.trim(input.val());
+
+    if (value) {
+      input.
+        addClass("has-items").
+        removeClass("is-empty");
+    } else {
+      input.
+        addClass("is-empty").
+        removeClass("has-items");
+    }
+  };
+
+  $("input[type='text']").
+    on("blur", markInput).
+    each(markInput);
+});
--- a/app/assets/stylesheets/components/_selectize-madlib.scss
+++ b/app/assets/stylesheets/components/_selectize-madlib.scss
@@ -1,12 +1,18 @@
 .madlib {
-  .selectize-control .selectize-input {
+  .selectize-control .selectize-input,
+  #{$all-text-inputs} {
     @include padding($smallest-spacing null);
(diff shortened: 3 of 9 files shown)


app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,45 @@
...
+  <%= form.input :problem,
+      label: t(".problem"),
+      placeholder: "Problem 4"%>
+
   <%= form.submit t(".add") %>
+  <%= link_to t("helpers.cancel_button"), :back %>
 <% end %>
+
+<script type = "text/javascript">
+  $(function() {
```
mborsare: This is straight from stackoverflow. Is there a better way to do it?

app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,45 @@
-<%= simple_form_for @assignment, url: manage_assessments_outcome_coverage_assignments_path do |form| %>
-  <%= form.input :name, label: t(".assignment") %>
-  <%= form.input :problem, label: t(".problem") %>
+<%= content_for :full_bleed do %>
```
mborsare: Should we turn this into a partial since we use it here and on the add a class/outcome form?

app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,45 @@
...
+          $(this).addClass("is-empty");
+          $(this).removeClass("has-items");
+        } else {
+          $(this).removeClass("is-empty");
+          $(this).addClass("has-items");
+        }
+      });
+    }
+
+    inp.blur(err);
```
derekprior: Do you want to do this only on blur, or would you like it done any time the user enters text (don't wait until they blur)?

app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,45 @@
-<%= simple_form_for @assignment, url: manage_assessments_outcome_coverage_assignments_path do |form| %>
-  <%= form.input :name, label: t(".assignment") %>
-  <%= form.input :problem, label: t(".problem") %>
+<%= content_for :full_bleed do %>
+  <div class="madlib-header">
+    <h1 class="headline-narrow">
+      Add assignment
+    </h1>
+    <h2 class="headline-narrow-subhead">
+      <%= @assignment.outcome_coverage.coverage.course.number %>
```
margaretoti: This can be made `assignment.outcome.course.name`. @derekprior, I am playing around in the console in an attempt to make this less brittle. Any ideas? I am happy to add some code in the model to accomplish this.

app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,45 @@
...
+      &middot;
+      <%= @assignment.outcome_coverage.coverage.course.name %>
```
margaretoti: `assignment.outcome.course.name`

app/views/manage_assessments/assignments/new.html.erb
derekprior: We would need a string of associations and delegates such that `assignment.course_number` works. 

something like:

```ruby
class Assignment
  has_one :coverage, through: :outcome_coverage
  delegate :course_number, to: :coverage
end

class Coverage
  delegate :number, to: :course, prefix: true
end
```

... should work.

app/views/manage_assessments/assignments/new.html.erb
derekprior: This is pretty close to how I'd do it, but we can clean it up a bit.

app/views/manage_assessments/assignments/new.html.erb
mborsare: Keeping it on blur is fine – unless there's some sort of performance issue.

app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,45 @@
...
+
   <%= form.submit t(".add") %>
+  <%= link_to t("helpers.cancel_button"), :back %>
 <% end %>
+
+<script type = "text/javascript">
+  $(function() {
+    var inp = $("input[type='text']");
+
+    function checkInput() {
```
mborsare: @derekprior I renamed this function, what do you think?

derekprior:
@mborsare I added commit with some refactoring of the JS, Ruby, and i18n. This looks good to me.

I didn't extract a template for the full bleed yet. Partials that accept blocks are always a little weird and hard to follow. A helper might make more sense. I might play with a bit in a different PR.

PR #189, opened 2017-05-25
Style madlib controls

mborsare:
This PR creates a partial that can be shared by all the madlib forms. It
also applies base styles to it.

screen shot 2017-05-25 at 11 28 57 am
screen shot 2017-05-25 at 11 29 04 am

https://trello.com/c/fAJ6Ijv4

### DIFF
--- a/app/views/manage_assessments/assignments/new.html.erb
+++ b/app/views/manage_assessments/assignments/new.html.erb
@@ -11,11 +11,9 @@
   </div>
 <% end %>
 
-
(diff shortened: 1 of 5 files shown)


app/views/manage_assessments/assignments/new.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= simple_form_for @assignment, url: manage_assessments_outcome_coverage_assignments_path do |form| %>
```
ryancoughlin: 🤗

PR #190, opened 2017-05-25
User can add attachments to an assignment

margaretoti:
[Trello card](https://trello.com/c/jv1d2jDD). This branch allows users to add multiple attachments to an assignment when creating an assignment. The attachment of files is handled using the [Paperclip gem](https://github.com/thoughtbot/paperclip). Currently users can only add files; they cannot edit or delete them. 

screen shot 2017-05-25 at 11 55 05 am

### DIFF
--- a/app/controllers/manage_assessments/assignments_controller.rb
+++ b/app/controllers/manage_assessments/assignments_controller.rb
@@ -8,7 +8,7 @@ class ManageAssessments::AssignmentsController < ApplicationController
   def create
     @assignment = outcome_coverage.build_assignment(assignment_params)
     authorize(@assignment)
-    
+
     if @assignment.save
       redirect_to manage_assessments_course_path(outcome_coverage.coverage.course),
         success: t(".success", label: outcome_coverage.outcome.label)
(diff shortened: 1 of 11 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 /log/*.log
 /tmp
 tmp
+/public
```
margaretoti: The test suite stored all files in this hugely nested folder.

Gemfile
```diff
@@ -13,6 +13,7 @@ gem "jquery-rails"
 gem "neat", "~> 1.7.0"
 gem "normalize-rails"
 gem "paper_trail"
+gem "paperclip", "~> 5.0.0"
```
margaretoti: @derekprior, I could not remember which version of Paperclip you updated to so I just copied the one in the Paperclip documentation. Happy to `bundle install` a different version.

.gitignore
derekprior: Some things in public we actually want. For instance, the error pages are stored in there. I think `public/system` should be good.

Gemfile
derekprior: This is good.

margaretoti:
@derekprior, Sounds good. I will wait for you to finish adding attachments to coverage. In the meantime, I might try to write a model validation to check that a user attaches only approved file formats (pdf for now).

PR #192, opened 2017-05-25
Allow adding attachments to coverages

derekprior:
When creating a coverage the user can optionally upload one or more
attachments. This UI is not what is reflected in the prototype, but the
functionality is in place. We can refine the UX in further iteration.

In dev and test, we use the file system for storage but use S3 for
staging and production. The S3 secret key is stored in encrypted secrets
storage.

MIT wants files uploaded here to remain private except to those with
appropriate access. For that reason, the files are uploaded to s3 with
private permissions. The URL for the file is set to a local path where
we will eventually implement our own controller that performs
authorization and redirects with an expiring URL to the actual file.
That part will be implemented in a future PR.


config/initializers/paperclip.rb
```diff
@@ -0,0 +1,23 @@
+Paperclip.interpolates :attachments_root do |attachment, style|
+  if attachment.options[:storage] == :filesystem
+    "public/attachments"
+  else
+    "attachments"
+  end
+end
+
+Paperclip.interpolates :coverage_id do |attachment, style|
+  attachment.instance.coverage.id
```
margaretoti: Why do you need `.instance` here?

derekprior: in this context, `attachment` means a wrapper around the actual file that is attached to our model. `instance` gets us from that back to the parent model instance (confusingly also called `Attachment` here...

margaretoti: Got it. Thanks. I was confusing it with our attachment.

PR #193, opened 2017-05-26
Design for a split tab component

mborsare:
This PR creates a component that allows the user to quickly go back and
forth between matched and unmatched outcomes.

split-tabs

https://trello.com/c/40JgUcWV

### DIFF
--- a/app/assets/javascripts/split_tabs.js
+++ b/app/assets/javascripts/split_tabs.js
@@ -1,0 +1,29 @@
+$(function() {
+  var splitTabsContainer = ".split-tabs";
+  var splitTab = ".split-tab";
+  var splitTabPanel = ".split-tab-panel";
+  var activeClass = "is-active";
+  var transitionTime = 300;
+
+  $(splitTab + ":first-child")
+    .addClass(activeClass)
+    .closest(splitTabsContainer)
+    .siblings(splitTabPanel + ":first-of-type")
+    .show();
+
+  $(splitTab).click(function() {
+    $(this)
+      .siblings(splitTab)
+      .removeClass(activeClass)
+      .closest(splitTabsContainer)
+      .siblings(splitTabPanel)
+      .hide();
+
+    $(this)
+      .addClass(activeClass)
+      .closest(splitTabsContainer)
+      .siblings(splitTabPanel)
+      .eq($(this).index(splitTab))
+      .fadeIn(transitionTime);
+  });
+});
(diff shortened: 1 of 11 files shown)


mborsare:
@derekprior and @margaretoti on Monday could one of you please help me with this? The matched and unmatched outcome totals are hard coded right now. It would be great to get that working.

derekprior:
@mborsare I added a commit to make those dynamic. JS looks fine to me. This looks good to me. I'd like to have @margaretoti take a look too.

Only comment I have is does it make sense to use the word "covered" rather than "matched"? Will that make more sense to the user?

db/schema.rb
```diff
@@ -16,164 +16,158 @@
   enable_extension "plpgsql"
 
   create_table "alignments", force: :cascade do |t|
-    t.integer  "outcome_id",          null: false
```
alexandra-aas: I would make sure these database changes are actually part of your PR.

app/assets/javascripts/split_tabs.js
```diff
@@ -0,0 +1,29 @@
...
+      .siblings(splitTabPanel)
+      .hide();
+
+    $(this)
+      .addClass(activeClass)
+      .closest(splitTabsContainer)
+      .siblings(splitTabPanel)
+      .eq($(this).index(splitTab))
+      .fadeIn(transitionTime);
+  });
```
alexandra-aas: I think we can simplify this a bit. I always try to avoid using `sibling` and `closest` because it relies on the markup not changing. One way to avoid this is to use `data-attributes`. For example, the matched tab could have `data-associated-panel: "matched"` and then the associated panel might have `data-panel: "matched"`. Then you might do the following jQuery:

```javascript
$(splitTab).click(function() {
  var associatedPanel = $(this).data("associated-panel");

  splitTabPanel.hide();
  $("[data-panel="+associatedPanel+"]").show();
});
```

Obviously there's a bit more to your code than just what I wrote above, but just wanted to give you a taste.

db/schema.rb
derekprior: They aren't. It's an artifact of a Rails 5.1 upgrade. I can do them separately and then rebase

app/assets/javascripts/split_tabs.js
derekprior: This is true. Not sure why I glossed over that in my review. Great point.

mborsare: Thanks @alexandra-aas, that's a great idea. I'm going to leave this a little brittle for now though for the sake of time. 

@derekprior please feel free to refactor in another PR. I was spending a little too much time on it this morning and figure you or @margaretoti can knock it out quicker than I.

PR #194, opened 2017-05-26
Clicking on an unmatched outcome redirects to a pre-populated form

margaretoti:
[Trello Card](https://trello.com/c/oGE82DIT)
This branch allows users on the `manage_assessments/courses/:course_id` page to add a single additional outcome to the listed course from this page. When a user clicks on an outcome, they are redirected to the `new_manage_assessments_course_coverages_path` with the selected icon already pre-populated. To accomplish this, the `outcome_id` is passed as a query parameter in the url to the `Coverages` controller. 

**Step One** - Click on Outcome A
screen shot 2017-05-26 at 2 47 23 pm

**Step Two** - Fill in pre-populated form
screen shot 2017-05-26 at 2 47 34 pm


spec/features/user_adds_coverage_to_a_course_spec.rb
```diff
@@ -22,9 +22,59 @@
     end
   end
 
+  scenario "user adds additional coverage to a course
```
margaretoti: Any advice on how to word this scenario? It is currently quite long.

app/controllers/manage_assessments/coverages_controller.rb
```diff
@@ -2,7 +2,12 @@ module ManageAssessments
   class CoveragesController < ApplicationController
     def new
       @coverage = course.coverages.build
-      @coverage.outcome_coverages.build
+      if outcome_id_given_in_query_params
```
derekprior: Do we need the conditional? What happens if we *always* do:

```
@coverage.outcome_coverages.build(outcome_id: params[:outcome_id])
```

If the param is `nil` it's no different than `build` with no params. Is there a problem if I pass an invalid id?

app/views/manage_assessments/coverages/new.html.erb
```diff
@@ -18,6 +18,7 @@
 
   <%= form.simple_fields_for :outcome_coverages do |fields| %>
     <%= fields.input :outcome_id,
+      selected: set_selected_attribute_with_outcome_id_if_present(@outcome),
```
derekprior: I believe this should happen automatically by virtue of the `outcome_id` already being set on the outcome coverage. What happens if you remove this and the the associated helper?

spec/features/user_adds_coverage_to_a_course_spec.rb
derekprior: the feature description already says something like "user adds coverage to a course" so we don't need to repeat that in the scenario. We could say something like:  `scenario "via unmatched outcome link"` or something like that.

spec/features/user_adds_coverage_to_a_course_spec.rb
```diff
@@ -22,9 +22,59 @@
...
+  from the coverages page by clicking on an unmatched outcome" do
+    unmatched_outcome = create(:outcome,
```
derekprior: I'd try to limit the setup code as much as possible here. I think maybe you can do something like:

```
course = create(:course)
matched_outcome, unmatched_outcome = create_pair(:outcome, course: course)
coverage = create(:coverage, course: course, outcomes: [matched_outcome])
user = user_with_assessments_access_to(course.department)
```

spec/features/user_adds_coverage_to_a_course_spec.rb
```diff
@@ -22,9 +22,59 @@
...
+    unselected_subject = create(:subject)
+    course = outcome.course
+    coverage = create(:coverage, course: course, outcomes: [outcome])
+    coverage.course.outcomes << unmatched_outcome # associate the unmatched_outcome with a course, but not a subject
+    user = user_with_assessments_access_to(course.department)
+
+    visit manage_assessments_course_path(course.id, as: user)
+    click_on unmatched_outcome.nickname
+
+    expect(page).to have_header("Add a class")
```
derekprior: I don't think this adds much to the test case.

spec/features/user_adds_coverage_to_a_course_spec.rb
```diff
@@ -22,9 +22,59 @@
...
+    expect(page).to have_prepopulated_select_box(unmatched_outcome.nickname)
```
derekprior: I think this is really all we care about in this test case. We're reasonably certain the form work because we tested it in the test case above. I think we can end this test case here. What do you think?

margaretoti: @derekprior, I agree and that was my first thought. I should have stuck with it. I will change it back now.

app/views/manage_assessments/coverages/new.html.erb
margaretoti: It almost works. The problem is that `selected` needs an `id` to pre-populate the field in the `new` form. If I only pass `@outcome`, the controller successfully redirects to the form but without the pre-populated outcome value. If I add `selected: @outcome.id`, the tests error saying `undefined method id for nil:NilClass`. I wrote the helper to get past this problem.

.gitignore
```diff
@@ -3,6 +3,7 @@
 /log/*.log
 /tmp
 tmp
+/public/system
```
derekprior: This seems to be left over from a previous branch and isn't needed.

PR #195, opened 2017-05-30
Refactor split tabs

derekprior:
The previous split-tab JavaScript was heavily dependent on the structure
of the markup. This implementation is instead dependent on being wired
up via data attributes.

The tab controls themselves must have a `data-tab` attribute set to some
identifier. The associated content must have a `data-content` with the
value set to the same value.

After wiring up the event handlers on the tabs, we trigger a click on
the first tab to set it as the active tab.

### DIFF
--- a/app/assets/javascripts/split_tabs.js
+++ b/app/assets/javascripts/split_tabs.js
@@ -1,29 +1,21 @@
 $(function() {
-  var splitTabsContainer = ".split-tabs";
-  var splitTab = ".split-tab";
-  var splitTabPanel = ".split-tab-panel";
   var activeClass = "is-active";
-  var transitionTime = 300;
+  var transitionTimeInMilliSeconds = 300;
 
-  $(splitTab + ":first-child")
-    .addClass(activeClass)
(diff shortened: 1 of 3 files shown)


mborsare:
Looks amazing!

app/assets/javascripts/split_tabs.js
```diff
@@ -1,29 +1,20 @@
...
+    otherTabControls.removeClass(activeClass);
 
-    $(this)
-      .addClass(activeClass)
-      .closest(splitTabsContainer)
-      .siblings(splitTabPanel)
-      .eq($(this).index(splitTab))
-      .fadeIn(transitionTime);
-  });
+    targetTabContent.fadeIn(300);
```
margaretoti: Is is possible to store `300` in a variable, similar to the original `var transitionTime = 300` (see old line 6 which is now in red) or the more descriptive `var transitionTimeInMilliSeconds = 300`?

derekprior: Yes. Good point.

PR #199, opened 2017-06-01
User can edit coverage for a course

margaretoti:
A user can now add an outcome to an existing coverage or change an existing outcome to a different outcome. To accomplish this functionality, `edit` and `update` where added to the `Coverages` Controller. A user cannot yet delete an outcome from a coverage. If a user tries to add a duplicate outcome to a coverage, she will be prompted to select another outcome.

**Add Another Outcome button added**
screen shot 2017-06-01 at 1 11 41 pm

**Edit form rendered with pre-populated fields**
screen shot 2017-06-01 at 1 12 01 pm

### DIFF
--- a/app/controllers/manage_assessments/coverages_controller.rb
+++ b/app/controllers/manage_assessments/coverages_controller.rb
@@ -6,6 +6,11 @@ module ManageAssessments
       authorize(@coverage)
     end
 
+    def edit
+      @coverage = Coverage.find(params[:id])
+      authorize(@coverage)
+    end
+
     def create
       @coverage = course.coverages.build(coverage_params)
       authorize(@coverage)
@@ -14,6 +19,17 @@ module ManageAssessments
         redirect_to manage_assessments_course_path(course)
       else
         render :new
+      end
+    end
+
+    def update
+      @coverage = Coverage.find(params[:id])
+      authorize(@coverage)
+
+      if @coverage.update_attributes(coverage_params)
+        redirect_to manage_assessments_course_path(@coverage.course)
+      else
+        render :edit
       end
     end
 
@@ -29,7 +45,7 @@ module ManageAssessments
         permit(
           :subject_id,
           attachments_attributes: [:id, :file, :_destroy],
-          outcome_coverages_attributes: [:coverage_id, :outcome_id],
+          outcome_coverages_attributes: [:id, :outcome_id],
         )
     end
   end
--- a/app/policies/coverage_policy.rb
+++ b/app/policies/coverage_policy.rb
@@ -2,4 +2,12 @@ class CoveragePolicy < ApplicationPolicy
   def create?
     user.manage_assessments?(record.course.department)
   end
+
+  def edit?
+    create?
+  end
+
+  def update?
+    create?
+  end
 end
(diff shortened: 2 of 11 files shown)


app/policies/coverage_policy.rb
```diff
@@ -2,4 +2,12 @@ class CoveragePolicy < ApplicationPolicy
   def create?
     user.manage_assessments?(record.course.department)
   end
+
+  def edit?
+    create?
```
margaretoti: `user.admin?(record.course.department)` could also be an option here.

app/policies/coverage_policy.rb
```diff
@@ -2,4 +2,12 @@ class CoveragePolicy < ApplicationPolicy
...
+  end
+
+  def update?
+    create?
```
margaretoti: and here as well.

app/controllers/manage_assessments/coverages_controller.rb
```diff
@@ -1,12 +1,19 @@
 module ManageAssessments
   class CoveragesController < ApplicationController
+    rescue_from ActiveRecord::RecordNotUnique, with: :outcome_not_unique
```
margaretoti: Should this instead be in `ApplicationsController`?

app/policies/coverage_policy.rb
derekprior: I think this is good.

app/views/manage_assessments/courses/_outcome_status.html.erb
```diff
@@ -4,7 +4,7 @@
     <h2 class="outcome-nickname"><%= outcome.nickname %></h2>
     <p class="outcome-card-description">
       <%= outcome.description %>
-    </p>
+    </p>
```
derekprior: Can you get rid of this trailing whitespace here? Nothing major, obviously. I have vim configured to do this for me automatically on save. You might be able to do similar with an atom package: https://github.com/atom/whitespace

config/locales/manage_assessments.en.yml
```diff
@@ -46,9 +46,11 @@ en:
         add: Add
         headline: Add assignment
     coverages:
+      coverage:
+        add_another_outcome: Add another outcome
       new:
-        add_attachment: Add an Attachment
-        add_outcome: Add an Outcome
+        add_attachment: Attach student work
```
derekprior: I think "Add an Attachment" was what was in the mockup when adding attachments to a subject (this is most likely a syllabus, not student work at this stage).

spec/models/outcome_coverage_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+describe OutcomeCoverage do
+  context "validations" do
+    subject do
```
derekprior: I have a personal preference to do this inside an "it" block rather than in the "subject" -- particularly since the subject really only applies to this one test case. You can see how I did this here: https://github.com/MIT-IR/abet/pull/201/files#diff-d05b36864e0c1f4fc61a91c251004dbeR4

I don't feel strongly about this though. If you prefer subject then I think that's fine here.

derekprior:
Looks good to me!

PR #200, opened 2017-06-01
Fix dynamic association insertion bugs

derekprior:
* We were using cocoon in an attempt to add associated attachments
inside a `ul`, but they were being added as siblings to the `ul`
instead. We can use data attributes to better control this through
cocoon.

* We were mistakenly running all cocoon insertions through the
attachment-specific JavaScript, which was resulting in some JavaScript
errors when trying to dynamically add outcomes to a coverage.


mborsare:
LGTM

PR #204, opened 2017-06-02
Various improvements to the assignment form

derekprior:
1. Now uses the madlib form builder with proper headlines as in the design.
2. Use the same JavaScript and partials for attachment handling as we use for coverage attachments.
3. Fix an issue rendering the name of the file when editing a coverage.


mborsare:
LGTM

PR #207, opened 2017-06-07
Run JavaScript feature specs in headless Chrome

derekprior:
Chrome 59 is now shipping, which supports headless operation. This PR
adds a new `headless_chrome` driver and sets Capybara to use it by
default for JavaScript specs. If it is useful during debugging to see
the tests function, you can change the driver to `chrome`.

### DIFF
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -1,5 +1,17 @@
+require "selenium/webdriver"
+
 Capybara.register_driver :chrome do |app|
   Capybara::Selenium::Driver.new(app, browser: :chrome)
 end
 
-Capybara.javascript_driver = :chrome
+Capybara.register_driver :headless_chrome do |app|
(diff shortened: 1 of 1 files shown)


spec/support/capybara.rb
```diff
@@ -1,5 +1,17 @@
+require "selenium/webdriver"
+
 Capybara.register_driver :chrome do |app|
   Capybara::Selenium::Driver.new(app, browser: :chrome)
 end
 
-Capybara.javascript_driver = :chrome
+Capybara.register_driver :headless_chrome do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: { args: %w(headless disable-gpu) }
```
tabfugnic: does `disable-gpu` help/is required on gpu-less systems?

derekprior: I meant to remove that from the options here, but I'm happy enough to leave it in. I don't want to use my GPU in tests.

PR #208, opened 2017-06-07
User can archive an outcome coverage

margaretoti:
When a user clicks Delete, the selected outcome coverage is archived. If a coverage only has one outcome coverage, the coverage and the outcome coverage are archived and no longer displayed on the UI. To accomplish this, an archived boolean column was added to both the `Outcome Coverages` and `Coverages` table. An `Archive` service object handles the archiving of objects.

### DIFF
--- a/app/models/course_coverage.rb
+++ b/app/models/course_coverage.rb
@@ -1,6 +1,6 @@
 class CourseCoverage < SimpleDelegator
   def has_coverages?
-    coverages.present?
+    coverages.where(archived: false).present?
   end
 
   def covered_outcomes_count
--- a/app/policies/outcome_coverage_policy.rb
+++ b/app/policies/outcome_coverage_policy.rb
@@ -1,0 +1,5 @@
+class OutcomeCoveragePolicy < ApplicationPolicy
+  def destroy?
+    user.manage_assessments?(record.coverage.course.department)
+  end
+end
--- a/spec/features/user_deletes_outcome_coverage_from_a_subject_spec.rb
+++ b/spec/features/user_deletes_outcome_coverage_from_a_subject_spec.rb
@@ -1,0 +1,37 @@
+require "rails_helper"
+
+feature "User deletes an outcome coverage for a subject" do
+  scenario "coverage has 2 or more outcome coverages" do
+    course = create(:course)
+    outcomes = create_pair(:outcome, course: course)
+    coverage = create(:coverage, course: course, outcomes: outcomes)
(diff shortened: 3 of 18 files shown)


app/policies/outcome_coverage_policy.rb
```diff
@@ -0,0 +1,5 @@
+class OutcomeCoveragePolicy < ApplicationPolicy
+  def destroy?
+    true
```
margaretoti: I need to change this. `user.manage_assessments?(record.outcome.course.department)`, perhaps?

spec/features/user_deletes_outcome_coverage_from_a_subject_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+feature "User deletes an outcome coverage for a subject" do
+  scenario "coverage has 2 or more outcome coverages" do
+    course = create(:course)
+    outcome_A, outcome_B = create_pair(:outcome)
```
derekprior: I think you can just do `outcomes = create_pair(:outcome)` as you don't ever need to refer to the outcomes indivudually here.

Also, should the outcomes be associated with the course you created in the line above?

spec/features/user_deletes_outcome_coverage_from_a_subject_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    visit manage_assessments_course_path(course.id, as: user)
+    first(".class-card-outcomes-wrapper").click_on("Delete")
+
+    expect(page).to have_outcome_number_of(1)
+  end
+
+  scenario "coverage has 1 outcome coverage with associated assignments" do
+    course = create(:course)
+    outcome = create(:outcome)
```
derekprior: Should this be associated with the course from the line above?

I think we could simplify a lot of this setup with some fancier factories, but that's out of scope for this PR.

app/models/course_coverage.rb
```diff
@@ -1,6 +1,6 @@
 class CourseCoverage < SimpleDelegator
   def has_coverages?
-    coverages.present?
+    coverages.where(archived: false).present?
```
margaretoti: Sadly, the `[courses_controller](https://github.com/MIT-IR/abet/blob/master/app/controllers/manage_assessments/courses_controller.rb)` depends on the `has_coverages` method and without specifying `archived: false`, I ran into problems.

PR #209, opened 2017-06-07
User can download an attachment

margaretoti:
A user can download an attachment from S3 from an expiring url. This url is not available to the general public. This functionality was accomplished by using Paperclip's [restricting access to objects on S3 features](https://github.com/thoughtbot/paperclip/wiki/Restricting-Access-to-Objects-Stored-on-Amazon-S3).  This branch also adds an attachment factory.

### DIFF
--- a/app/controllers/manage_assessments/attachments_controller.rb
+++ b/app/controllers/manage_assessments/attachments_controller.rb
@@ -1,0 +1,10 @@
+module ManageAssessments
+  class AttachmentsController < ApplicationController
+    def show
+      @attachment = Attachment.find(params[:id])
+      authorize(@attachment)
+
+      redirect_to @attachment.expiring_url
+    end
+  end
+end
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -1,4 +1,6 @@
 class Attachment < ActiveRecord::Base
+  EXPIRE_TIMEFRAME = 60.seconds
+
   belongs_to :attachable, polymorphic: true
 
   has_attached_file :file,
@@ -10,5 +12,17 @@ class Attachment < ActiveRecord::Base
 
   def name
     file_file_name
+  end
+
+  def expiring_url
+    file.expiring_url(Time.now + EXPIRE_TIMEFRAME)
+  end
+
+  def course_department
+    attachable.course_department
+  end
+
+  def subject_department
+    attachable.subject_department
   end
 end
--- a/app/policies/attachment_policy.rb
+++ b/app/policies/attachment_policy.rb
@@ -1,0 +1,5 @@
+class AttachmentPolicy < ApplicationPolicy
+  def show?
+    user.manage_assessments?(record.course_department)
+  end
+end
--- a/spec/controllers/manage_assessments/attachments_controller_spec.rb
+++ b/spec/controllers/manage_assessments/attachments_controller_spec.rb
@@ -1,0 +1,17 @@
+require "rails_helper"
+
+describe ManageAssessments::AttachmentsController do
+  describe "GET #show" do
+    it "downloads the requested attachment from server" do
+      assignment = create(:assignment)
+      attachment = create(:attachment, attachable: assignment)
+      department = assignment.course_department
+      user = user_with_assessments_access_to(department)
+
+      sign_in(user)
+      get :show, params: { id: attachment.id }
+
+      expect(response).to redirect_to(attachment.expiring_url)
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -19,6 +19,8 @@ Rails.application.routes.draw do
 
     resources :outcome_coverages, only: [:destroy]
 
+    resources :attachments, only: [:show]
+
     resources :assessments, only: [:new, :create, :edit, :update] do
       resource :archive, only: [:create, :destroy]
     end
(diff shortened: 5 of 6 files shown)


app/controllers/manage_assessments/attachments_controller.rb
```diff
@@ -0,0 +1,10 @@
+module ManageAssessments
+  class AttachmentsController < ApplicationController
+    def show
+      @attachment = Attachment.find(params[:id])
+      authorize(@attachment)
+
+      redirect_to @attachment.file.expiring_url(Time.now + 60.seconds)
```
margaretoti: Currently using an arbitrary time in the `redirect_to`.

app/policies/attachment_policy.rb
```diff
@@ -0,0 +1,5 @@
+class AttachmentPolicy < ApplicationPolicy
+  def show?
+    user.manage_assessments?(record.attachable.outcome_coverage.coverage.course.department)
```
margaretoti: Any suggestions on how to make this less brittle?

derekprior: We should set up a chain of delgations so you can do something like: `record.attachable.department`

This gets a little cmplicated because coverages themselves can have two different departments: The department of the course that added the coverage or the department of the subject. Those don't have to be the same thing. I'm dealing with this a bit in the work I'm doing for results.

I think it might make sense to do: `Coverage#course_department` and `Coverage#subject_department` so we are explicit about these things.

Then you'd have to set up methods/delegations so you could do `record.attachable.course_department`. Does this make sense?

app/controllers/manage_assessments/attachments_controller.rb
derekprior: I think this timeout is okay, but I see that we're repeating it a couple of times between the test and implementation. We're also having to jump through `.file` in both locations.

What do you think about adding `expiring_url` as a method directly on `Attachment`, that would call `file.expiring_url(Time.now + EXPIRE_TIMEFRAME)` (a constant we would set on the class). I think that would let us keep this logic contained and also keep us from having to do `attachment.file.`

spec/controllers/manage_assessments/attachments_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+describe ManageAssessments::AttachmentsController do
+  describe "GET #show" do
+    it "downloads the requested attachment from server" do
+      course = create(:course)
+      outcome = create(:outcome)
+      coverage = create(:coverage, course: course, outcomes: [outcome])
+      outcome_coverage = OutcomeCoverage.first
+      assignment = create(:assignment, outcome_coverage: outcome_coverage)
```
derekprior: I have simplified some of this setup with factories in my results branch which may make sense to include here to clean this up. Try these out:

```ruby
  factory :assignment do
    name
    problem "problem 1"

    after(:build) do |assignment|
      if assignment.outcome_coverage.nil?
        coverage = build(:coverage)
        assignment.outcome_coverage = coverage.outcome_coverages.first
      end
    end
  end
```

```ruby
  factory :coverage do
    transient do
      outcomes []
    end

    course
    subject

    after(:build) do |coverage, evaluator|
      if evaluator.outcomes.present?
        coverage.outcomes = Array(evaluator.outcomes)
      else
        coverage.outcomes = [
          build(:outcome, course: coverage.course)
        ]
      end
    end
  end
```

If you do this, I think you'll be able to say `create(:assignment)` and get an assignment where the various relationships all talk about the same things.

Do you want to try these out and see if they work for you?

spec/controllers/manage_assessments/attachments_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      attachment = Attachment.create(file: File.open("spec/fixtures/attachments/example.pdf"), attachable: assignment)
```
derekprior: I do think this would make for a useful factory. Then you could `create(:attachment, attachable: assignment)` and we could default to using the example pdf.

app/models/attachment.rb
```diff
@@ -11,4 +13,16 @@ class Attachment < ActiveRecord::Base
   def name
     file_file_name
   end
+
+  def expiring_url
+    file.expiring_url(Time.now + EXPIRE_TIMEFRAME)
+  end
+
+  def course_department
+    self.attachable.course_department
```
derekprior: I don't think you need the `self` on these. 

You could also do this with a `delegates` call, I think:

```
delegates :course_department, :subject_department, to: :attachable
```

That's what we've done in other places. Having said that, I think I might actually prefer writing the methods out as you have done here. It seems more clear. What do you think?

spec/support/capybara.rb
```diff
@@ -14,4 +14,4 @@
     desired_capabilities: capabilities
 end
 
-Capybara.javascript_driver = :headless_chrome
+Capybara.javascript_driver = :chrome
```
derekprior: Have you updated to Chrome 59? If so, I think this shouldn't be necessary. If you're still getting that error after updating, then certainly leave it.

app/models/attachment.rb
margaretoti: I like the methods. They seem clearer to me as well.

spec/support/capybara.rb
margaretoti: Good point. It does work now that Chrome has been updated to Version 59.

PR #210, opened 2017-06-08
Report results on assignments

derekprior:
We want to record results against assignments rather than assessments.
Assessments are being replaced by assignments. This change updates the
models to accept results on assignments, but it leaves the old fields
and relationships in place. We still need to migrate existing data and
then we should be able to clean up old data and set, for example,
`assignment_id` on `results` to `null: false`.


app/models/assignment.rb
```diff
@@ -1,16 +1,26 @@
 class Assignment < ActiveRecord::Base
   belongs_to :outcome_coverage
 
+  has_one :outcome, through: :outcome_coverage
+  has_one :coverage, through: :outcome_coverage
+  has_one :subject, through: :coverage
+  has_one :course, through: :coverage
+
   has_many :attachments, as: :attachable
+<<<<<<< HEAD
```
margaretoti: Accidental rebase comments. I think you want only this bit. 
```
accepts_nested_attributes_for :attachments,
      allow_destroy: true,
      reject_if: :all_blank
```

derekprior: Thanks!

PR #215, opened 2017-06-13
Update assessments dashboard content

derekprior:
* Fix the progress bar so it reports outcomes covered/total outcomes
* Link or redirect to course coverage page as appropriate
* Update language from assessments -> coverage


app/controllers/manage_assessments/dashboard_controller.rb
```diff
@@ -1,10 +1,24 @@
 class ManageAssessments::DashboardController < ApplicationController
+  before_action :redirect_to_course_coverage, if: :has_access_to_one_course?
```
margaretoti: I like this. Much cleaner.

PR #222, opened 2017-06-14
Order course outcomes in the controller

margaretoti:
This branch uses a method on the Outcome class to order outcomes by label name, instead of ordering them via the `has_many outcomes` relation on the Course model.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -3,7 +3,7 @@ class Course < ActiveRecord::Base
 
   has_many :coverages, -> { where archived: false }
   has_many :outcome_coverages, through: :coverages
-  has_many :outcomes, -> { order(:label) }
+  has_many :outcomes
 
   def self.without_outcomes
     where(outcomes_count: 0)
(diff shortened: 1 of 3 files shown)


app/controllers/manage_outcomes/outcomes_controller.rb
```diff
@@ -4,6 +4,7 @@ class ManageOutcomes::OutcomesController < ApplicationController
 
   def index
     @course = course
+    @outcomes = course.outcomes.order(:label)
```
derekprior: this is a case where I'd consider adding a method to the model so you can avoid adding another instance variable in the controller. You could create `Outcomes#alphabetical` and then do `@course.outcomes.alphabetical` in the view.

Another thing people do sometimes is add an association like: `has_many :alphabetical_outcomes, class_name: 'Outcome', -> { order(:label) }`. This would allow you to use eager loading to eager load the association in its sorted order. I don't think we need to do that here for performance reasons, but I thought I'd mention it as a possibility just in case.

I don't feel strongly about the ivar vs `alphabetical` scope on the model. What do you think?

PR #225, opened 2017-06-15
Add additional activities to the Activity Feed

margaretoti:
This branch adds paper_trail ([documentation here](https://github.com/airblade/paper_trail)) to assignments and outcome coverages. The activity feed now displays when the following have been created or updated: 
* Results
* Assignments
* Outcome Coverages

Clicking the link associated with a result directs the user to the `manage_results_assignment_path`, where the user can view and edit the result. Clicking the link associated with an assignment directs the user to the corresponding course page (`manage_assessments_course_path`) for that assignment. 

activity_feed1

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -1,17 +1,11 @@
-class Activity < PaperTrail::Version
-  default_scope { where(item_type: "Result") }
-
-  belongs_to :user, foreign_key: :whodunnit
-  belongs_to :result, foreign_key: :item_id
+class Activity < ApplicationRecord
+  belongs_to :user
+  belongs_to :item, polymorphic: true
 
   delegate :username, to: :user, allow_nil: true
 
   def date
     created_at.to_date
-  end
-
-  def subject
-    result.assignment.subject.to_s
   end
 
   def time_formatted
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -18,4 +18,10 @@ class Assignment < ActiveRecord::Base
   delegate :department, to: :course, prefix: true
 
   validates :name, presence: true
+
+  has_paper_trail
+
+  def subject_name
+    subject.to_s
+  end
 end
--- a/app/models/outcome_coverage.rb
+++ b/app/models/outcome_coverage.rb
@@ -7,4 +7,10 @@ class OutcomeCoverage < ActiveRecord::Base
   delegate :label, :nickname, to: :outcome, prefix: true
 
   validates :outcome_id, uniqueness: { scope: [:coverage_id, :archived] }, unless: :archived
+
+  has_paper_trail
+
+  def subject_name
+    coverage.subject.to_s
+  end
 end
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -23,6 +23,10 @@ class Result < ActiveRecord::Base
     "#{assessment_name} - #{assessment_description}"
   end
 
+  def subject_name
+    assignment.subject.to_s
+  end
+
   private
 
   def denormalize_department
--- a/app/policies/activity_policy.rb
+++ b/app/policies/activity_policy.rb
@@ -2,8 +2,7 @@ class ActivityPolicy < ApplicationPolicy
   class Scope < Scope
     def resolve
       scope.
-        joins(result: :department).
-        where(departments: { slug: user.department_slugs }).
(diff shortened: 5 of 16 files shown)


app/models/assignment.rb
```diff
@@ -18,4 +18,10 @@ class Assignment < ActiveRecord::Base
   delegate :department, to: :course, prefix: true
 
   validates :name, presence: true
+
+  has_paper_trail
+
+  def subject_name
+    "#{subject.number} - #{subject.title}"
```
derekprior: This is the definition of `subject.to_s`. What do you think of calling that here instead?

app/models/outcome_coverage.rb
```diff
@@ -7,4 +7,10 @@ class OutcomeCoverage < ActiveRecord::Base
   delegate :label, :nickname, to: :outcome, prefix: true
 
   validates :outcome_id, uniqueness: { scope: [:coverage_id, :archived] }, unless: :archived
+
+  has_paper_trail
+
+  def subject_name
+    "#{coverage.subject.number} - #{coverage.subject.title}"
```
derekprior: Same question here.

app/models/result.rb
```diff
@@ -23,6 +23,10 @@ def name
     "#{assessment_name} - #{assessment_description}"
   end
 
+  def subject_name
+    "#{assignment.subject.number} - #{assignment.subject.title}"
```
derekprior: Same here...

app/views/activities/_activity.html.erb
```diff
@@ -9,16 +9,27 @@
...
   <td class="activity-event-and-subject">
     <span class="activity-event">
-      <%= t(
-        ".event.#{activity.event}_html",
-        assignment: link_to(
-          activity.result.name,
-          manage_results_assignment_path(activity.result.assignment_id)
-        ),
-      ) %>
+      <% if activity.item_type == "Result" %>
```
derekprior: What do you think of breaking these conditionals out into their own partials? So we could do something like:

```
render "#{activity.item_type.downcase}_activity", item: item
```

app/models/activity.rb
```diff
@@ -1,8 +1,10 @@
 class Activity < PaperTrail::Version
-  default_scope { where(item_type: "Result") }
+  default_scope { where(item_type: ["Result", "Assignment", "OutcomeCoverage"]) }
 
   belongs_to :user, foreign_key: :whodunnit
   belongs_to :result, foreign_key: :item_id
+  belongs_to :assignment, foreign_key: :item_id
+  belongs_to :outcome_coverage, foreign_key: :item_id
```
derekprior: These end up being a lie in some cases, right? For instance, if I get an instance of an activity for an Assignment, I can call `activity.result` and either get nil (if there's no result with that same id) or a result that just so happens to have the same id as an assignment.

The `belongs_to :result` worked before because we constrained it so they were all result activities. I think if we expand this model to cover more types, we have to embrace the existing polymorphic `item` association to get at the thing the activity points to.

app/policies/activity_policy.rb
```diff
@@ -2,8 +2,6 @@ class ActivityPolicy < ApplicationPolicy
   class Scope < Scope
     def resolve
       scope.
-        joins(result: :department).
-        where(departments: { slug: user.department_slugs }).
```
derekprior: This is what limits the activities to being for departments you have access to. This will be hard to do now that we want to show different types of activities. My first thought was we could say `joins(item: :department)` and then ensure assignment, result, and coverage all define a `department` association. However, this won't work because `item` is a polymorphic association and we can't call `joins` with it.

I can't come up with a great way to overcome this right now.... Hmmm....

config/locales/en.yml
```diff
@@ -36,10 +36,26 @@ en:
...
+        update_assignment_html: Assignment %{name} updated for %{subject}
+    outcome_coverage_activity:
+      event:
+        create_outcomecoverage_html: Outcome coverage created for %{subject}
+        update_outcomecoverage_html: Outcome coverage updated for %{subject}
+    result_activity:
+      event:
+        create_result_html: Result created for %{assignment} in %{subject}
+        update_result_html: Result updated for %{assignment} in %{subject}
+    # activity:
```
derekprior: delete me :)

PR #227, opened 2017-06-19
Allow users to undo dynamic "add outcome"

derekprior:
When adding outcomes to a subject, users can click "Add outcome" to add
another outcome. If they decide against doing this, they can now click
"undo" to remove the dynamically added form fields for this.

The "undo" link is only shown for freshly-added inputs. You cannot
remove or delete an existing outcome from a subject. Instead, you can
archive it, which is accomplished from the list of subjects.


mborsare:
👍

PR #232, opened 2017-06-20
Disallow editing existing coverages

derekprior:
Consider the scenario where a user:

1. Adds outcomes to subject 1.00
2. Adds assignments to those outcomes
3. Clicks "Add an outcome" to add another outcome to this subject

Previous to this change, while adding additional outcomes the user could
also change the subject or edit the previous coverage to point at
different coverages. This could potentially cause assignments, results,
and attachment to become disconnected from the actual outcome they were
intended for.

Now, when a user elects to "Add an outcome" they can only add new
outcomes. They cannot change the subject or any previously existing
outcomes.

If the user wants to remove an outcome from a subject, they should use
the `Delete` link on the listing page.

### DIFF
--- a/app/controllers/manage_assessments/coverages_controller.rb
+++ b/app/controllers/manage_assessments/coverages_controller.rb
@@ -8,6 +8,7 @@ module ManageAssessments
 
     def edit
       @coverage = Coverage.find(params[:id])
+      @coverage.outcome_coverages.build
       authorize(@coverage)
     end
 
@@ -26,7 +27,7 @@ module ManageAssessments
       @coverage = Coverage.find(params[:id])
       authorize(@coverage)
 
-      if @coverage.update_attributes(coverage_params)
+      if @coverage.update_attributes(update_coverage_params)
         redirect_to manage_assessments_course_path(@coverage.course)
       else
         render :edit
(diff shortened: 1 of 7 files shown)


app/controllers/manage_assessments/coverages_controller.rb
```diff
@@ -45,8 +46,12 @@ def coverage_params
         permit(
           :subject_id,
           attachments_attributes: [:id, :file, :_destroy],
-          outcome_coverages_attributes: [:id, :outcome_id, :_destroy]
+          outcome_coverages_attributes: [:outcome_id, :_destroy]
         )
     end
+
+    def update_coverage_params
+      coverage_params.except(:subject_id)
```
margaretoti: When I mentored with Paul, he mentioned not liking the `except` keyword because if an attribute is added to `coverage_params` it has to be explicitly excluded from `update_coverage_params`, which is easy to forgot. Although not nearly as clear, is it better to go with the following? 
```
def update_coverage_params
    params.require(:coverage).permit(
    attachments_attributes: [:id, :file, :_destroy],
    outcome_coverages_attributes: [:outcome_id, :_destroy]
)
```

app/controllers/manage_assessments/coverages_controller.rb
```diff
@@ -45,8 +46,12 @@ def coverage_params
         permit(
           :subject_id,
           attachments_attributes: [:id, :file, :_destroy],
-          outcome_coverages_attributes: [:id, :outcome_id, :_destroy]
+          outcome_coverages_attributes: [:outcome_id, :_destroy]
         )
     end
+
+    def update_coverage_params
```
margaretoti: When I mentored with Paul, he mentioned not liking the `except` keyword because if an attribute is added to `coverage_params` it has to be explicitly excluded from `update_coverage_params`, which is easy to forgot. Although not nearly as clear, is it better to go with the following? 
```
def update_coverage_params
    params.require(:coverage).permit(
    attachments_attributes: [:id, :file, :_destroy],
    outcome_coverages_attributes: [:outcome_id, :_destroy]
)
```

derekprior: Thats a fair point that I had not considered...

PR #234, opened 2017-06-20
Add confirmation when archiving outcome coverage

derekprior:
This is recoverable, but (currently) only with help from Jon or a
developer. Let's give them a quick "are you sure?" message...

abet_outcomes_assessment


mborsare:
Great idea!

PR #237, opened 2017-06-20
User can add a result to an assignment

margaretoti:
A user can now add a result to an assignment from the `manage_assessments_courses_path` using the Add Result link. This link can be used to add multiple results across semesters and years.


app/views/manage_assessments/outcome_coverages/_outcome_coverage.html.erb
```diff
@@ -34,6 +34,7 @@
               edit_manage_assessments_outcome_coverage_assignment_path(
                 outcome_coverage_id: outcome_coverage.id,
                 assignment: outcome_coverage.assignment) %>
+        <%= link_to "Add Result", new_manage_results_assignment_result_path(outcome_coverage.assignment) %>
```
derekprior: Should we move this to i18n?

margaretoti: Good call.

PR #238, opened 2017-06-20
Archived outcomes shown in unmatched outcomes list

margaretoti:
On the `manage_assessments_courses_path`, archived outcomes now appear in the list of unmatched outcomes as clickable. Previously, archived outcomes were displayed as crossed-out meaning that the user could not re-add an archived outcome by selecting that outcome under the Unmatched Outcomes tab.

### DIFF
--- a/app/models/outcome.rb
+++ b/app/models/outcome.rb
@@ -4,7 +4,7 @@ class Outcome < ActiveRecord::Base
   belongs_to :course, counter_cache: true
 
   has_one :department, through: :course
-  has_many :outcome_coverages
+  has_many :outcome_coverages, -> { where archived: false }
   has_many :alignments
   has_many :standard_outcomes, through: :alignments
 


app/helpers/outcomes_helper.rb
```diff
@@ -1,6 +1,6 @@
 module OutcomesHelper
   def outcome_card_class(outcome)
-    if outcome.outcome_coverages.present?
+    if outcome.unarchived_outcome_coverages.present?
```
derekprior: What do you think about making this a "feature" of the `Outcome#outcome_coverages` association, just as we did with `Coverage#outcome_coverages`?

I think as written here, it's going to cause an N+1 query because each outcome in that view is going to filter the outcome coverages.

app/models/outcome.rb
```diff
@@ -4,7 +4,7 @@ class Outcome < ActiveRecord::Base
   belongs_to :course, counter_cache: true
 
   has_one :department, through: :course
-  has_many :outcome_coverages
+  has_many :outcome_coverages, -> { where archived: false }
```
margaretoti: @derekprior, Very true. It also means that I only need to change one file.

PR #240, opened 2017-06-20
Disable HTML5 required validation on MadLib forms

derekprior:
We now have nice custom styling on errors for madlib forms. HTML5
validations (of which we were only using the `required` validations) are
nice when they work, but they don't notify the user of any problems on
our `selectize` select boxes because the underlying input is actually
hidden.

We can let these hit the server validation instead so the user gets
better error messages. To facilitate this, I added the `novalidate`
attribute to the form tag for our madlib forms.

This change also includes some validation fixes so errors that the
server encounter are rendered properly.

* Scope the exception to uniqueness validations appropriately. We were
previously excepting all validations on those fields when we only wanted
to except the uniqueness validation.
* Add a user visible validation to the first blank outcome_coverage when
a coverage has no outcomes.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -30,7 +30,10 @@ module ApplicationHelper
   end
 
   def madlib_form_for(path, options = {}, &block)
-    options = options.deep_merge(html: { class: :madlib }, wrapper: :madlib)
+    options = options.deep_merge(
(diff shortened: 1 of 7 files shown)


derekprior:
abet_outcomes_assessment

derekprior:
@margaretoti I added a test for this. Good idea.

PR #243, opened 2017-06-21
Activity feed links auto scroll

margaretoti:
The activity feed reports when an assignment has been created or updated, and includes a link to the specific assignment. Now when a user clicks on an assignment link, she will be directed to the point on the courses page that lists the assignment. To accomplish this feature, an anchor is passed to the `manage_assessments_course_path` when the activity type is assignment. The `id` of that assignment is added to the html on the courses page.


derekprior:
:+1:

PR #244, opened 2017-06-21
Rename Manage Assessments tab

margaretoti:
The Manage Assessments tab has been renamed to Manage Assignments tab.


derekprior:
:+1:

PR #247, opened 2017-06-21
Rename assessments to assignments

margaretoti:
This branch changes `assessments` to `assignments` across the app. These changes touch routes, namespaced controllers, models, permissions, policies, variable names in spec files. Migrations and .scss files have not yet been changed.


app/models/permission.rb
```diff
@@ -1,12 +1,12 @@
 class Permission
   ADMIN = "CAN EDIT ABET"
-  ASSESSMENTS = "CAN EDIT ASSESSMENTS"
+  ASSIGNMENTS = "CAN EDIT ASSIGNMENTS"
```
derekprior: "CAN EDIT ASSESSMENTS" is how the permission is set in the external system we get permissions from. We may want to rename the constant but I think we need to keep the string value it's set to for integration purposes.

PR #248, opened 2017-06-21
Show tab bar when user has no coverages yet

derekprior:


derekprior:
This doesn't work right with the red background change. I think it's not worth addressing right now.

PR #251, opened 2017-06-22
Remove choose file button from edit assignment

margaretoti:
[Trello card](https://trello.com/c/P0dYb3tH)
This branch removes the choose file button present on the edit
assignment page.

### DIFF
--- a/app/assets/stylesheets/components/_add-item.scss
+++ b/app/assets/stylesheets/components/_add-item.scss
@@ -68,3 +68,7 @@
     text-align: right;
   }
 }
+
+.is-hidden {
+  display: none;
+}
--- a/app/views/attachments/_fields.html.erb
+++ b/app/views/attachments/_fields.html.erb
@@ -1,6 +1,6 @@
 <li class="attachment">
   <%= f.hidden_field :id %>
-  <%= f.file_field :file, as: :file %>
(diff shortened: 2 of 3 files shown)


app/views/attachments/_fields.html.erb
```diff
@@ -1,5 +1,5 @@
 <li class="attachment">
-  <%= f.file_field :file, as: :file %>
+  <%= f.file_field :file, as: :file, style: "display: none" %>
```
mborsare: Hmm, I'm wondering if an inline style is the most elegant way to achieve this?

derekprior: Is there a utility class for "hidden" that we should use?

mborsare: There isn't one currently.

margaretoti:
@derekprior, I believe so. @mborsare, Mind taking one more look.

app/assets/stylesheets/components/_add-item.scss
```diff
@@ -68,3 +68,7 @@
     text-align: right;
   }
 }
+
+.remove_default_choose_file_button {
```
mborsare: I wonder if since this is just one declaration we could create a utility class instead? What do you think of something like...

```
.is-hidden {
  display: none;
}
```

PR #252, opened 2017-06-22
Always show remove attachment link

derekprior:
If the user cancels a file attachment dialog, they previously had no way
to remove the empty file input. It wasn't necessary to do so, but I can
definitely see confusion resulting from that.

Ideally there would be an event we could handle when the user cancels
that dialog, but there is not. No good workarounds exist for this
either. As a stopgap solution, this change removes the code that was
keeping the "delete" link hidden until there was a file added.

Now the user can at least manually clean up after themselves.

### DIFF
--- a/app/assets/javascripts/attachments.js
+++ b/app/assets/javascripts/attachments.js
@@ -1,16 +1,13 @@
 $(function() {
   var fileInput = "input[type='file']";
   var fileName = ".file-name";
-  var removeLink = "a.remove_fields";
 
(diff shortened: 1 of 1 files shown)


derekprior:
abet_outcomes_assessment

mborsare:
If this is the best solution then I'm fine with it. Like you said, at least you can manually clean it up if you want. Perhaps there's a copy change we can make down the line to clear things up.

PR #253, opened 2017-06-22
Add navigation from result to course page

margaretoti:
This branch adds a link from a newly created result for an assignment
to the courses page that the assignment belongs. 

screen shot 2017-06-22 at 3 10 26 pm

### DIFF
--- a/app/assets/stylesheets/components/_headline-with-actions.scss
+++ b/app/assets/stylesheets/components/_headline-with-actions.scss
@@ -1,12 +1,11 @@
 .headline-with-actions {
+  align-items: center;
   display: flex;
+  margin-bottom: $large-spacing;
 
   h1 {
     flex: 1;
+    margin-bottom: 0;
     margin-right: $small-spacing;
-  }
-
-  .button {
-    align-self: flex-start;
   }
 }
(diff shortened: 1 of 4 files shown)


app/views/manage_results/assignments/show.html.erb
```diff
@@ -5,6 +5,7 @@
 
   <div>
     <% if policy(@assignment).create_results? %>
+      <%= link_to "Back to #{@assignment.course.name}", manage_assignments_course_path(@assignment.course) %>
```
mborsare: Does the "Back to" bit deserve a translation?

mborsare: This feels like it needs some padding. It's very close to the button.

app/assets/stylesheets/components/_headline-with-actions.scss
```diff
@@ -1,12 +1,11 @@
 .headline-with-actions {
   display: flex;
```
mborsare: I know I added these lines. So it's my fault. But we like to order our properties within a declaration block alphabetically. So it'd be ...

```
align-items: ...;
display: ...;
margin-bottom: ...;
```

mborsare:
LGTM

PR #254, opened 2017-06-22
Display input errors inline

mborsare:
We previously weren't showing any error messages. This adds the message
to display "under the line" on the madlib forms.

uh-oh


app/assets/stylesheets/components/_madlib-statement.scss
```diff
@@ -16,3 +16,27 @@
...
+.field_with_errors {
+  .selectize-control .selectize-input.items {
+    background-color: tint($required-color, 90%);
+    border-bottom-color: $required-color;
+  }
+
+  .error {
+    display: block;
+
+    @at-root .error {
```
ryancoughlin: wat dis?

app/assets/stylesheets/components/_madlib-statement.scss
```diff
@@ -16,3 +16,27 @@
...
+      background: image-url("icon_error.svg") 100% center / 1.35em no-repeat;
+      color: $required-color;
+      display: none;
+      font-size: $small-font-size;
+      padding-right: 2em;
```
ryancoughlin: You and ali are king/queen of variables on padding variables.....................should this be a variable?

app/assets/javascripts/input_validation.js
```diff
@@ -0,0 +1,5 @@
+$(function(){
+  $(document).on("change", ".field_with_errors", function() {
```
ryancoughlin: Interesting, does this get triggered when the DOM of `.field_with_errors` changes?

app/assets/stylesheets/components/_madlib-statement.scss
mborsare: It tells sass, "hey bump this up outside of the nested selector and make a rule." It's a way to keep things a little more organized. So, rather that having a `.error { display: none}` randomly in the sheet. Now, there's a logical relationship.

app/assets/javascripts/input_validation.js
mborsare: I'm pretty sure, yes. @derekprior wrote this. He can confirm.

app/assets/stylesheets/components/_madlib-statement.scss
mborsare: This is one of those cases where no variable felt right. Much like the type scale conversation we had the other day.

PR #255, opened 2017-06-26
Render all attachments above the "add" button

derekprior:
Previous to this change, pre-existing attachments would render above the
"Add and Attachment" button while new attachments would render below.
They now all render above the button, which is a sibling list element to
the attachments themselves.

I also updated the `fields_for` call so it doesn't render a hidden
`input` between the `li` tags for the attachment id, which is not valid
HTML. Instead, we manually render the `id` inside the `li`.

### DIFF
--- a/app/views/attachments/_fields.html.erb
+++ b/app/views/attachments/_fields.html.erb
@@ -1,4 +1,5 @@
 <li class="attachment">
+  <%= f.hidden_field :id %>
   <%= f.file_field :file, as: :file %>
 
   <% if f.object.persisted? %>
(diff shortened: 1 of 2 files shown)


derekprior:
abet_outcomes_assessment_and_slack_-_thoughtbot

app/models/user.rb
```diff
@@ -15,4 +15,8 @@ def permissions
   def departments
     Department.where(slug: department_slugs)
   end
+
+  def foo
```
margaretoti: I don't think that you need this function.

derekprior: HA!  Left over debugging from a blog post I was writing. Thanks for catching this!

PR #257, opened 2017-06-26
Display username on activity feed

margaretoti:
This branch displays the user responsible for the activity listed on the Activity Feed. To accomplish this feature, `before_action :set_paper_trail_whodunnit` has been added to the `Applications Controller`. As of PaperTrail 5.0, this line is no longer a default of PaperTrail ([see change logs](https://github.com/airblade/paper_trail/blob/master/CHANGELOG.md)).

screen shot 2017-06-26 at 2 08 33 pm

### DIFF
--- a/spec/features/user_views_activity_feed_spec.rb
+++ b/spec/features/user_views_activity_feed_spec.rb
@@ -34,4 +34,24 @@ feature "User views activity feed" do
                                     assignment: other_department_result.assignment.name,
                                     subject: other_department_result.assignment.subject)
   end
+
+  scenario "sees her username is associated with each activity" do
+    outcome = create(:outcome)
+    course = outcome.course
+    coverage = create(:coverage, course: course, outcomes: [outcome])
+    outcome_coverage = OutcomeCoverage.first
+    user = user_with_assignments_access_to(course.department)
+    user_creates_an_assignment(outcome_coverage, user)
+
+    visit activities_path(as: user)
+
+    expect(page).to have_content user.username
+  end
+
+  def user_creates_an_assignment(outcome_coverage, user)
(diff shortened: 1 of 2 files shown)


app/models/activity.rb
```diff
@@ -12,6 +12,11 @@ def time_formatted
     created_at.strftime("%-I:%M %p")
   end
 
+  def whodunnit
+    user_id = PaperTrail.whodunnit
```
margaretoti: I cannot think of a scenario where PaperTrail.whodunnit would be `nil`. Can you? If so, I will add a conditional to deal with what to do when `nil`

app/models/activity.rb
```diff
@@ -12,6 +12,11 @@ def time_formatted
     created_at.strftime("%-I:%M %p")
   end
 
+  def whodunnit
```
derekprior: Do we need this `whodunnit` bit? We should be accessing the user record by the `User` association on `Activity`. In the SQL view, we map `whodunnit` to `user_id` to make this work.

spec/features/user_views_activity_feed_spec.rb
```diff
@@ -34,4 +34,24 @@
...
+    user = user_with_assignments_access_to(course.department)
+    user_creates_an_assignment(user, course)
+
+    visit activities_path(as: user)
+
+    expect(page).to have_content user.username
+  end
+
+  def user_creates_an_assignment(user, course)
+    visit manage_assignments_course_path(course.id, as: user)
```
derekprior: can we speed this up a tad by going right to `new_manage_assignments_outcome_coverage_assignment_path` (wow, that's a mouthful) rather than navigating to the course first?

PR #258, opened 2017-06-26
Typography tweaks for adding and editing an assignment intro

mborsare:
This design is more holistic. The type styles now echo those of the blank
state for a course.

screen shot 2017-06-26 at 3 15 03 pm
screen shot 2017-06-26 at 3 14 57 pm

https://trello.com/c/KR5D5uBL

### DIFF
--- a/app/assets/stylesheets/components/_madlib-leadin.scss
+++ b/app/assets/stylesheets/components/_madlib-leadin.scss
@@ -1,0 +1,8 @@
+.madlib-leadin {
+  margin-bottom: $base-spacing;
+  padding-left: $smallest-spacing;
+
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/components/_madlib-leadin.scss
```diff
@@ -0,0 +1,9 @@
+.madlib-leadin {
+  color: $medium-gray;
```
conchan: just checking to see if it makes sense to set a style/variable/mixin for texts of this hierarchal level

mborsare: Good call – I already have a class I can reuse

PR #260, opened 2017-06-27
Remove unused layout block

derekprior:
The `header_with_filters` block was used in the old assessments filter
header which we don't have anymore.


mborsare:
🔥

PR #267, opened 2017-06-28
Require results to have an assignment id

derekprior:
Now that the historical data has been imported, all existing results
have an assignment id. We can now mark this as required.

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -4,8 +4,7 @@ class Result < ApplicationRecord
 
   before_save :denormalize_department
 
-  belongs_to :assessment, counter_cache: true
-  belongs_to :assignment, required: true
+  belongs_to :assignment
   belongs_to :department
 
   validates :percentage, presence: true, inclusion: { in: 0..100 }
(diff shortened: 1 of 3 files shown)


app/models/result.rb
```diff
@@ -4,8 +4,7 @@ class Result < ApplicationRecord
 
   before_save :denormalize_department
 
-  belongs_to :assessment, counter_cache: true
-  belongs_to :assignment, required: true
+  belongs_to :assignment
```
margaretoti: Should `required: true` remain on this line? Perhaps, we no longer need it as it is the default in Rails 5. I do remember you saying in the past that because we updated from Rails 4, we have to be explicit. It is possible that that only applies for models added before the Rails 5.0 upgrade.

derekprior: `required: true` adds a model validation. In this commit, we change to use a database constrain instead. A model validation isn't particularly useful here because if the user somehow manages to submit a request to create a result that doesn't have an associated assignment, then we've probably got a bug somewhere and a validation error message wouldn't help.

I wrote an article about this not too long ago: https://robots.thoughtbot.com/validation-database-constraint-or-both

PR #269, opened 2017-06-28
Moving the delete control for an outcome

mborsare:
This creates a stronger association with what is being deleted (the
outcome) and the control.

delete


derekprior:
:+1:

PR #271, opened 2017-06-29
User can view and delete attachments

margaretoti:
This branch allows the user to view and delete attachments from the courses page. Clicking on the Attachments link located just below the outcome description expands and collapses the list of attachments associated with an assignment. If no attachments belong to an assignment, the Attachments link will instead read Attach student work. Clicking on Attach student work will redirect the user to `assignment#edit`.·

screen shot 2017-06-29 at 9 59 49 am

### DIFF
--- a/app/views/manage_assignments/outcome_coverages/_outcome_coverage.html.erb
+++ b/app/views/manage_assignments/outcome_coverages/_outcome_coverage.html.erb
@@ -36,6 +36,15 @@
 
     <div class="class-card-assignment-controls">
       <% if outcome_coverage.assignment.present? %>
+        <% if outcome_coverage.assignment.attachments.present? %>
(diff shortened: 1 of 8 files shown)


app/views/manage_assignments/outcome_coverages/_outcome_coverage.html.erb
```diff
@@ -34,6 +34,15 @@
 
     <div class="class-card-assignment-controls">
       <% if outcome_coverage.assignment.present? %>
+        <% if outcome_coverage.assignment.attachments.present? %>
```
margaretoti: I probably should put this in a partial.

margaretoti:
@derekprior, Thanks for taking a look. I think the tests relate to failures with chrome as well because they all pass locally. I will rebase and run the tests again.

PR #272, opened 2017-06-29
Pluralize "Matched Outcome" appropriately

derekprior:
Before this change, a subject that had a single outcome would display "1
Matched Outcomes". Now it will correctly say "1 Matched Outcome".


mborsare:
👍

PR #273, opened 2017-06-29
User can attach a syllabus from courses page

margaretoti:
A user can now attach a syllabus from the `manage_assignments_courses_path`. The button Attach a syllabus will redirect the user to `coverage#edit`. If a syllabus has already been attached, the file name will appear as a downloadable link. 

screen shot 2017-06-29 at 6 20 20 pm

### DIFF
--- a/app/views/manage_assignments/coverages/_coverage.html.erb
+++ b/app/views/manage_assignments/coverages/_coverage.html.erb
@@ -13,7 +13,15 @@
     </h2>
     <p class="class-card-heading-outcomes">
       <%= t(".matched_outcomes", count: coverage.outcome_coverages.size) %>
-    </p>
+      <div>
+        <% if coverage.attachments.present? %>
+          <%= link_to coverage.attachment.name,
+            manage_assignments_attachment_path(coverage.attachment) %>
+        <% else %>
+          <%= link_to t(".attach-syllabus"),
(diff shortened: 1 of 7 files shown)


app/views/manage_assignments/coverages/_coverage.html.erb
```diff
@@ -13,6 +13,9 @@
     </h2>
     <p class="class-card-heading-outcomes">
       <%= t(".matched_outcomes", count: coverage.outcome_coverages.size) %>
+    <p>
```
derekprior: Should this only show if there are no attachments already? If there's already an attachment, should we just show a link to it?

Or should this always be a link and we can link to the existing attachments in the form?

margaretoti: This is a good point. @mborsare, What do you think? I lean having a link to Add a syllabus only if there currently is not one. Else, there is a link to the attachment.

mborsare: I agree with you @margaretoti

mborsare: I'm assuming one attachment at the class level though. Perhaps that's misguided?

margaretoti: @mborsare, You are correct the user can only attach one item at the class level.

PR #274, opened 2017-06-29
Design for collapsible attachment component

mborsare:
This PR establishes base styles for attachments on class cards.

attachments

Related to:
https://trello.com/c/d6Yu2iOA

### DIFF
--- a/app/views/manage_assignments/courses/_attachments.html.erb
+++ b/app/views/manage_assignments/courses/_attachments.html.erb
@@ -1,14 +1,11 @@
-<table class="assignment-attachments is-hidden">
-  <tbody>
-    <% attachments.each do |attachment| %>
-      <tr>
-        <td>
-          <%= link_to attachment.name, manage_assignments_attachment_path(attachment) %>
-        </td>
-        <td>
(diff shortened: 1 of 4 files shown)


app/views/manage_assignments/courses/_attachments.html.erb
```diff
@@ -1,12 +1,10 @@
...
-      <tr>
-        <td><%= attachment.file_file_name %></td>
-        <td><%= link_to t(".delete-link"),
-              manage_assignments_attachment_path(attachment),
-              method: :delete %></td>
-      </tr>
-    <% end %>
-  </tbody>
-</table>
+<ul class="assignment-attachments is-hidden attachments">
```
alexandra-aas: I wonder if these classes should go `attachments assignment-attachments is-hidden`. I realize that's a super nitpick, but I usually like to read things as general to specific. Maybe they are in that order currently?

mborsare: @alexandra-aas Interesting point. I usually just start tacking things on in no particular order. My thought is that even if I do order them by specificity the next person to touch the code will have to deduce that. It's not obvious solely from the markup.

PR #277, opened 2017-06-29
User views attachment number

margaretoti:
This branch adds the number of attachments associated with an attachment next to the attachments link.

screen shot 2017-06-29 at 4 45 48 pm

### DIFF
--- a/app/views/manage_assignments/outcome_coverages/_outcome_coverage.html.erb
+++ b/app/views/manage_assignments/outcome_coverages/_outcome_coverage.html.erb
@@ -39,7 +39,10 @@
         <%= inline_svg "paper_clip.svg", class: "add-item-icon" %>
 
         <% if outcome_coverage.assignment.attachments.present? %>
-          <%= link_to t(".attachments-expandable-link"), "#", class: "attachments-expandable-link class-card-assignment-control" %>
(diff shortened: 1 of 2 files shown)


app/views/manage_assignments/outcome_coverages/_outcome_coverage.html.erb
```diff
@@ -37,7 +37,10 @@
     <div class="class-card-assignment-controls">
       <% if outcome_coverage.assignment.present? %>
         <% if outcome_coverage.assignment.attachments.present? %>
-          <%= link_to t(".attachments-expandable-link"), "#", class: "attachments-expandable-link" %>
+          <%= link_to "#", class: "attachments-expandable-link" do %>
+            <%= t(".attachments-expandable-link") %>
+            <span><%= outcome_coverage.assignment.attachments.count %></span>
```
derekprior: Calling `.count` on an association will *always* trigger a database query to get a count of records. This can become a performance problem.

Calling `.length` will always count the number of items loaded into the array in memory. This is performant, but can be the wrong thing to do if all you really need is the count of records (i.e. you do not need to get the actual attachments, just the count).

There's a third option, `.size` which will act like `length` if the association is already loaded but act like `count` if it's not. That's *generally* what I'd use in these situations, but you could make an argument for `length` here instead.

PR #280, opened 2017-07-03
Design for the attachment count component

mborsare:
This PR extends an existing component to better fit into its context. It
also DRYs up some of the border declarations throughout.

attachments

https://trello.com/c/tSoQIx2O

### DIFF
--- a/app/assets/stylesheets/components/_split-tabs.scss
+++ b/app/assets/stylesheets/components/_split-tabs.scss
@@ -40,8 +40,20 @@
   font-weight: $bold;
   margin: $smallest-spacing;
   padding: 0.125em 1em;
+  transition: $base-transition;
+
+  a:hover &,
+  a:hover + & {
+    background-color: $mit-red;
+  }
 
   .is-active & {
     background-color: $mit-red;
+  }
+
+  &.emphasized-tab-digit-small {
+    @include padding(null 0.85em);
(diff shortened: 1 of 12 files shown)


alexandra-aas:
From a visual perspective, I wonder if on hover it would look a bit cleaner if the underline just went under the text and then the  background of the count bubble changed to maroon.

app/assets/stylesheets/components/_split-tabs.scss
```diff
@@ -45,3 +45,9 @@
     background-color: $mit-red;
   }
 }
+
+.emphasized-tab-digit-small {
```
alexandra-aas: If this is a modified of `.split-tab-digit`, I would nest it inside `.split-tab-digit` to show that it needs to be coupled with it to work.

alexandra-aas: If this is a modified of `.split-tab-digit`, I would nest it inside `.split-tab-digit` to show that it needs to be coupled with it to work.

mborsare:
Good idea! I'll give it a shot.
```
From a visual perspective, I wonder if on hover it would look a bit cleaner if the underline just went under the text and then the background of the count bubble changed to maroon
```

mborsare:
attachments-clean

PR #282, opened 2017-07-05
View attachments associated with coverages

margaretoti:
This branch allows the user to view all attachments associated with a
coverage. By clicking Add a syllabus, the user can attach a file or
files associated with the entire coverage. The expected user behavior is
that each coverage has only one document (a syllabus). From the
expandable view the user can download the current attachments, delete
them, or add another attachment.

screen shot 2017-07-05 at 4 23 00 pm

### DIFF
--- a/spec/features/user_deletes_attachment_from_a_coverage_spec.rb
+++ b/spec/features/user_deletes_attachment_from_a_coverage_spec.rb
@@ -1,0 +1,36 @@
+require "rails_helper"
+
+feature "User views attachment count and deletes an attachment from
+  a coverage", js: true do
+  scenario "successfully" do
+    coverage = create(:coverage)
+    attachment = create(:attachment, attachable: coverage)
+    attachment_count = coverage.attachments.size
+    subject_name = coverage.subject.title
+    user = user_with_assignments_access_to(coverage.course.department)
+
+    visit manage_assignments_course_path(coverage.course.id, as: user)
+
+    within(".class-card-assignment-control") do
+      expect(page).to have_content(attachment_count)
+    end
+
+    click_on t("manage_assignments.outcome_coverages.outcome_coverage.attachments-expandable-link")
+
+    delete_attachment
+
+    expect(page).to have_content(subject_name)
+    expect(page).to have_no_content(attachment.file_file_name)
+    expect(page).to have_content t("manage_assignments.coverages.coverage.attach-syllabus")
+
+    within(".class-card-heading") do
+      expect(page).to have_no_content t("manage_assignments.outcome_coverages.outcome_coverage.attachments-expandable-link")
+    end
+  end
+
+  def delete_attachment
+    within(".assignment-attachments") do
+      click_on t("manage_assignments.shared.attachments.delete_link")
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


spec/features/user_deletes_attachment_from_courses_page_spec.rb
```diff
@@ -27,4 +26,31 @@
       expect(page).to have_no_content(attachment_count)
     end
   end
+
+  scenario "related to a specific coverage successfully" do
```
mxie: What are your thoughts on moving this out as its own spec? I think it would be good to keep this feature spec focused on courses only.

margaretoti: Thanks @mxie. I will do that now.

spec/features/user_deletes_attachment_from_a_coverage_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    within(".class-card-assignment-control") do
+      expect(page).to have_content(attachment_count)
+    end
+
+    click_on t("manage_assignments.outcome_coverages.outcome_coverage.attachments-expandable-link")
+
+    within(".assignment-attachments") do
+      click_on t("manage_assignments.shared.attachments.delete_link")
+    end
```
mxie: Is there a better way to describe what the interaction is here? I know that we're clicking on a link/button within a certain part of the page, but it's not entirely clear as to what's going on from a glance.

I'm thinking something like the following:

```ruby
scenario "..." do
  # ...
  delete_assignment
end

def delete_assignment
  within ... do
    click_on ...
  end
end
```

spec/features/user_deletes_attachment_from_a_coverage_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    subject_name = coverage.subject.title
+    user = user_with_assignments_access_to(coverage.course.department)
+
+    visit manage_assignments_course_path(coverage.course.id, as: user)
+
+    within(".class-card-assignment-control") do
+      expect(page).to have_content(attachment_count)
+    end
+
+    click_on t("manage_assignments.outcome_coverages.outcome_coverage.attachments-expandable-link")
```
mxie: Same re: replacing this interaction with a method that better describes the interaction.

spec/features/user_deletes_attachment_from_a_coverage_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    within(".assignment-attachments") do
+      click_on t("manage_assignments.shared.attachments.delete_link")
+    end
+
+    expect(page).to have_content(subject_name)
+    expect(page).to have_no_content(attachment.file_file_name)
+
+    within(".class-card-heading") do
```
mxie: Hmm...I'm not sure if there's any value in verifying that the expand link is still there. Can you explain why this might be useful?

margaretoti: Sure. The expand link should disappear if no attachments are present. In its, place there should be a button to attach a syllabus. I will add an expectation that that button is present. @mxie

mborsare:
attachments

